PR #3, opened 2020-09-17
Webapp

MateoLostanlen:
The goal of this PR is to propose a webApp to allow anyone to use a pyronear predictor without any installation


MateoLostanlen:
@frgfm Hi ! When I try to install the package 'pyronearEngine' and then import 
`from PyronearEngine.inference.pyronearPredict import PyronearPredictor` it fail. However the install seems to work correctly, there is somthing wrong in the setup.py file ?

frgfm:
> @frgfm Hi ! When I try to install the package 'pyronearEngine' and then import
> `from PyronearEngine.inference.pyronearPredict import PyronearPredictor` it fail. However the install seems to work correctly, there is somthing wrong in the setup.py file ?

Mmmh what kind of error did you get?
Did you reinstall the package after making the changes to `PyronearPredictor`?

MateoLostanlen:
Yes I reinstalled the package and I got the following error : `ModuleNotFoundError: No module named 'pyronearEngine.inference'`

frgfm:
> Yes I reinstalled the package and I got the following error : `ModuleNotFoundError: No module named 'pyronearEngine.inference'`

Oh so you just need to edit `pyronearEngine/__init__.py` accordingly!

MateoLostanlen:
It works, thanks :)

pyronearEngine/inference/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) Pyronear contributors.
+# This file is dual licensed under the terms of the CeCILL-2.1 and AGPLv3 licenses.
+# See the LICENSE file in the root of this repository for complete details.
+
+from .pyronearPredict import *
```
frgfm, line 7: if you do `import *` , then please add `__all__ = ['PyronearPredictor']` after the imports of pyronearEngine/inference/pyronearPredict.py (it will limit the imported objects)

frgfm (changes requested):
Just a little change to clean import structure and we're good to go!

pyronearEngine/inference/__init__.py
MateoLostanlen: ok I made a classic import it is simpler considering there is only one import here

frgfm (approved):
Thanks for the edits!
We'll need to add some tests for the portions of the code that are not dependent on raspberry pi (pyronearPredict for instance), but we can do this in another PR!

MateoLostanlen:
Ok let's open an issue for that topic

PR #4, opened 2020-09-21
feat:create class to monitor pi system

fe51:
This feature aims to monitor some metrics from raspberry pi system.

In order to design a suitable system, I think some metrics could be recorded.
Then, after test phases, especially if the system is not connected to the internet, the output can be analyzed to understand how to improve the system or understand why it crashed.
As an example, in order to know if fans, heatsink are needed, cpu temperature can be recorded; Or, some performance about cpu usage or memory available could help to select the best raspberry pi model fitting our needs.

Output of this feature is a csv, an example is provided in the following zip file [pi_perf.csv.zip](https://github.com/pyronear/PyronearEngine/files/5258112/pi_perf.csv.zip) and below

datetime | cpu_temperature_C | mem_available_GB | cpu_usage_percent
-- | -- | -- | --
01/08/20 12:26 | 57.94 | 3.00 | 0.0
01/08/20 12:26 | 56.97 | 3.00 | 0.3
01/08/20 12:26 | 57.45 | 3.00 | 0.2
01/08/20 12:26 | 56.97 | 3.00 | 0.2


MateoLostanlen:
Hello @fe51, please can you check your conflict ?

fe51:
@MateoLostanlen it was mainly because of renaming PR #8. As long as requirements are specific to the engine I have created a specific requirement file for raspberryPi

MateoLostanlen (approved):
Hi @fe51, I finally tested your PR (longest review of all time). Everything seems to work. There are some problems with the package name between pyronearEngnine and pyroengine but that makes sense. I'll change that in a future PR in which I'll reorganize the repo.

PR #8, opened 2020-11-11
refactor: Renamed package

frgfm:
This PR updates package name and updates CI cache.

If I missed something, please let me know pray


pyroengine/__init__.py
```diff
@@ -5,4 +5,3 @@
 # See the LICENSE file in the root of this repository for complete details.
 
 from .version import __version__
-from pyronearEngine import inference
```
fe51, line 8: You delete the line because it was useless ? Otherwise, we should just change `pyronearEngine ` to `pyroengine` ?

fe51 (approved):
Thanks for the PR.

exept one question, it seems okay to me !

pyroengine/__init__.py
frgfm: Well, not useless but it meant:
`from pyroengine import *` would import everything declared in ̀inference`

I thought that `from pyronear.inference import ...` is a better & cleaner interface.

MateoLostanlen (approved):
All good :)

PR #9, opened 2020-12-03
Update Pyronear predictor

MateoLostanlen:
The purpose of this PR is to propose an improvement of the PyronearPredictor. The goal is to make it easier to adapt and to configure using a simple configuration file.


pyroengine/inference/configManager.py
```diff
@@ -0,0 +1,43 @@
...
+                            'use_center_crop': True}
+
+    with open(filename, 'w') as configfile:
+        config.write(configfile)
+
+
+if __name__ == '__main__':
+
+    # Run this module to save your config
+    write_config_file()
```
frgfm, line 43: Why make it a script?

pyroengine/inference/configManager.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
frgfm, line 1: for a python package, it would be better to have snake case naming of file :pray:

pyroengine/inference/inference.cfg
```diff
@@ -0,0 +1,13 @@
+[MODEL]
+backbone = rexnet1_0x
+num_classes = 1
+use_adaptive_concat_pool2d_head = True
+nb_features = 1280
+cut = -2
+checkpoint = pyro_checkpoint_V0.1.pth
```
frgfm, line 7: Shouldn't the checkpoint either be passed by the user (depending on local path) or be a remote publicly accessible URL?

pyroengine/inference/pyronearPredict.py
```diff
@@ -1,41 +1,93 @@
...
+# This file is dual licensed under the terms of the CeCILL-2.1 and AGPLv3 licenses.
+# See the LICENSE file in the root of this repository for complete details.
+
 import torch
 from torch import nn
 from PIL import Image
 from torchvision import transforms
+from pyroengine.inference.configManager import read_config_file
+from pyrovision.models.utils import cnn_model
+import holocron
```
frgfm, line 13: requirements need to be updated I think

frgfm (commented):
Thanks for the PR! While keeping the same spirit, I think that holocron dep + archi definition should be done in pyrovision (and add the checkpoint in release attachments). That way, in pyro-inference we can simply import the model that we want and pass `pretrained=True` to load the latest weights

MateoLostanlen:
@frgfm thanks for your review. I am closing this PR, to move it into pyro-vision. This way the last model always comes from the source repo without having to modify it in the others. I will add your remarks to the new PR

PR #10, opened 2020-12-03
Pc Pyronear Engine

MateoLostanlen:
The purpose of this PR is to propose a pc version of the Pyronear Engine. Mainly for test or demo purposes.


MateoLostanlen:
not relevent due to #13

PR #11, opened 2021-02-21
Re setup repo properly

MateoLostanlen:
The goal of this PR is to redo the setup of this repo properly by reproducing what has been done on pyrovision. In this PR the following points have been modified or created:

- Add docs
- Update licence
- Update README
- Add CONTRIBUTING file
- Update github workflow
- update setup & requierements
- add conlect_env script


MateoLostanlen:
These tests do not pass for the moment because it is necessary to install pyrovision which is not yet available on pip

fe51 (approved):
Hey ! Thanks fo structuring the repo !
Sounds good to me !  
Do you have any idea when pyro-vision will be available on pip ?

MateoLostanlen:
We need to make a new realese for that, maybe it's time to do it, @frgfm  what do you think ?

frgfm:
@MateoLostanlen my bad I didn't have access to a laptop the past few days! 
Still need to take care of a few things before the release, but I'll try to finish this by the end of the weekend

PR #13, opened 2021-02-21
refactor: Reorganized package

MateoLostanlen:
The purpose of this PR is to reorganize the repo. Now we have:
 - a pyronear_predictor class that loads the last pytonear model and makes a prediction on a given image.
- A Pyronear Engine class that manages the whole process of wildfire detection, from the image loading to the api call.
- A class that inherits from pyro Engine for each platform with an adapted image acquisition method. 

Pyro Engine functioning is currently simplistic, it should be improved in a future PR. Here we launch an alert each time we detect fire (so every n second when a fire outbreak), this is obviously not possible in reality.


docs/source/engine.rst
```diff
@@ -0,0 +1,24 @@
+pyroengine.engine
+=================
+
+The models subpackage contains everuthing to manage the whole Fire Detection process by capturing and saving the image and
```
frgfm, line 4: typo: everything

docs/source/engine.rst
```diff
@@ -0,0 +1,24 @@
...
+by predicting if there is a fire or not based on this image
```
frgfm, line 5: predicting whether there is a fire on this image

docs/source/engine.rst
```diff
@@ -0,0 +1,24 @@
...
+
+
+.. currentmodule:: pyroengine.engine
+
+Pyronear Predictor
+------------------
+
+The pyronear_predictor class that loads the last pytonear model and makes a prediction on a given image.
```
frgfm, line 13: typo here : PyronearPreditor

pyroengine/engine/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+from .pyronear_predictor import PyronearPredictor
+from .pyronear_engine import PyronearEngine
```
frgfm, line 7: let's rename the files "predictor.py" and "engine.py" to avoid complicated names (sine the package is already named pyroengine)

pyroengine/engine/pyronear_engine.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+from .pyronear_predictor import PyronearPredictor
+from pyroclient import client
+
+
+api_url = "http://pyronear-api.herokuapp.com"
```
frgfm, line 10: I think this needs to be loaded from environment variable using a dotenv file for security, and maintenance

docs/source/engine.rst
```diff
@@ -0,0 +1,24 @@
...
+
+.. autoclass:: PyronearPredictor
+
+
+Pyronear Engine
+---------------
+
+The Pyronear Engine class that manages the whole process of wildfire detection, from the image loading to the api call.
```
frgfm, line 21: I think you can remove this since there is almost the same thing in the docstring, so it will be displayed

docs/source/engine.rst
frgfm: Actually , you can even remove the whole line considering the docstring will be displayed in the documentation

pyroengine/engine/pyronear_predictor.py
```diff
@@ -3,45 +3,37 @@
...
         # Model definition
-        self.model = torchvision.models.resnet18()
-
-        # Change fc
-        in_features = getattr(self.model, 'fc').in_features
-        setattr(self.model, 'fc', nn.Linear(in_features, 2))
-
-        self.model.load_state_dict(torch.load(checkpointPath))
+        self.model = rexnet1_0x(pretrained=True)
+        self.model.eval()
```
frgfm, line 22: `self.model = rexnet1_0x(pretrained=True).eval()`

pyroengine/platforms/pc/pc_engine.py
```diff
@@ -0,0 +1,44 @@
...
+    Example
+    -------
+    # For a prediction every 5 frame
+    python pi_engine.py api_login, api_password 5
+    """
+    def run(self, every_n_frame=10):
+
+        cap = cv2.VideoCapture(0)
+
+        frame_nb = 0
```
frgfm, line 26: This might overflow at some point. I suggest have a frame count that is reset everytime a frame is analyzed

pyroengine/platforms/raspberry/pi_engine.py
```diff
@@ -0,0 +1,45 @@
...
+                    self.predict(frame)
+                    time.sleep(sleep_time)
+            finally:
+                camera.stop_preview()
+
+
+if __name__ == "__main__":
+    # Get inputs
+    api_login = sys.argv[1]
+    api_password = sys.argv[2]
```
frgfm, line 40: this should not be passed by the user (not sure but I think this does expect script arguments from the user)

pyroengine/engine/pyronear_engine.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def run(self):
+        """Should be implemented for each patform, with an adapted image capture method"""
+        pass
+
+    def predict(self, frame):
+
+        res = self.pyronearPredictor.predict(frame)
+        if res > 0.5:
+            print('Fire detected !!!')
```
frgfm, line 38: A few things:
- let's switch from console printing to logging
- perhaps updating the message itself would be better `f"Wildfire detection ({res:.2%})"`

pyroengine/engine/pyronear_engine.py
```diff
@@ -0,0 +1,55 @@
...
+            frame.save('fire.jpg')
```
frgfm, line 39: We lose precious time here. Currently:
- if a wildfire is detected, the loaded image will be saved to disk
- then the api will load it from disk and upload it with the API client

A better option:
- if a wildfire is detected, pass the loaded image directly to send the alert (rather than writing + loading from disk)
- if needed, also save the image to disk

frgfm (changes requested):
Thanks a lot for the PR! I left many comments, I hope it's not too much haha
Let me know what you think

MateoLostanlen:
@frgfm thanks for the review, I made the changes. Juste one thing should I include the .env file to the repo (changing the api credential) ?

frgfm:
> @frgfm thanks for the review, I made the changes. Juste one thing should I include the .env file to the repo (changing the api credential) ?

Oh my bad, I don't know how I missed the notification!
We do not push any .env file to the repo, but we can add each secret as a Github secret fortunately! More secure option :ok_hand: Only contributors need to have a copy of the .env file

MateoLostanlen:
I'm removing the test for know, we need to setup a raspbery pi env in an upcomming PR.

The broken link is the docs, it will be fix with this PR

the errors in codacy are on files that do not exist anymore ...

fe51 (approved):

frgfm:
@MateoLostanlen but all the tests were passing, there was only one failing because of an import to be renamed :sweat_smile:

MateoLostanlen:
@frgfm  Yes there was an import problem I just fixed it. But this does not change the problem, the import of picamera just bug after because we are not in a raspberry environment as you can see. Can I delete the tests again?

(well you will see, when https://github.com/pyronear/pyro-vision/pull/114 will be merge :) )

frgfm:
@MateoLostanlen let me patch something quickly so that this gets solved by https://github.com/pyronear/pyro-vision/pull/115 :)

frgfm:
I just made bud fix release to address this:
- Github release: https://github.com/pyronear/pyro-vision/releases/tag/v0.1.2
- pypi release: https://pypi.org/project/pyrovision/

I suggest we update pyro-engine requirements to `pyrovision>=0.1.2` in another PR to avoid issues for users with legacy versions

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 sphinx<3.5.0
 sphinx-rtd-theme==0.4.3
 pyrovision
+python-dotenv >= 0.15.0
```
frgfm, line 4: I didn't notice before : these are not a documentation requirements, both these should be moved to https://github.com/pyronear/pyro-engine/blob/master/requirements.txt

(I suggest replacing `pyrovision` with `pyrovision>=0.1.2` while you are at it)

.github/workflows/main.yml
```diff
@@ -31,41 +31,41 @@ jobs:
...
+            key: ${{ runner.os }}-pkg-deps-${{ hashFiles('requirements.txt') }}-${{ hashFiles('**/*.py') }}
+            restore-keys: |
+              ${{ runner.os }}-pkg-deps-${{ hashFiles('requirements.txt') }}-
+              ${{ runner.os }}-pkg-deps-
+              ${{ runner.os }}-
+        - name: Install dependencies
+          run: |
+            python -m pip install --upgrade pip
+            pip install -e .
+            pip install -r .github/workflows/requirements.txt
```
frgfm, line 56: Strange, did you add spaces? :thinking:

MateoLostanlen:
@frgfm thanks for the Release, I will make the new PR to update pyrovision version tonight

pyroengine/engine/predictor.py
```diff
@@ -0,0 +1,37 @@
...
+                                      normalize
+                                      ])
+
+    def predict(self, im):
+        imT = self.tf(im)
+
+        with torch.no_grad():
+            pred = self.model(imT.unsqueeze(0))
+
+        return torch.sigmoid(pred).item()
```
frgfm, line 37: I suggest moving the `torch.sigmoid` in the no_grad context just for safety ;)

MateoLostanlen:
there is a missing dependency in this PR (dotenv), I need to fix that too

frgfm (changes requested):
A few things to fix and we're ready to merge!

.github/workflows/main.yml
MateoLostanlen: just an indentation

frgfm:
> @frgfm thanks for the Release, I will make the new PR to update pyrovision version tonight

If you do, on top of the version in `setup.py`, please update the `README.md` badge :pray:

MateoLostanlen:
All good :)

.github/workflows/main.yml
```diff
@@ -30,43 +30,6 @@ jobs:
...
-          coverage run -m unittest discover test/
-          coverage xml
-
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
-        with:
-          file: ./coverage.xml
-          flags: unittests
-          fail_ci_if_error: true
-
```
frgfm, line 69: I thought they were passing with our previous fix?

frgfm (commented):
I'm not sure whether it was by mistake but the CI job for unittest has been removed again?

.github/workflows/main.yml
MateoLostanlen: haha yes beacause it fail due to picamera module, do you whant to see it ?

MateoLostanlen: https://github.com/pyronear/pyro-engine/pull/13/checks?check_run_id=2219038010 :)

MateoLostanlen: At the end we could remove picamera from the package because pyro-engine will not manage the capture anymore

frgfm: Alright, let's comment the test file and not the CI job then ! (otherwise we don't get a proper coverage)

frgfm: The CI won't have a camera ever so we need to find a way to test the other features but we'll do this in another PR!

MateoLostanlen: At the end the cameras will be managed via a simple script on the pi_zero. We will take this script out of the package and will no longer depend on picamera

frgfm: ? what do you mean? picture taking should be done in python since we do inference with it as well, shouldn't it?
(Perhaps we need to discuss the higher purpose of this repo in an issue, I'm really getting lost here :sweat_smile: It is a refactoring but if we don't know towards what organization we refactor, we might be overlooking many things)

MateoLostanlen:
closing this for #17

PR #15, opened 2021-03-23
Local api for api devices based on Fastapi 

fe51:
This PR aims at initiating a skeleton of an internal api for devices on sites using [Fastapi](https://fastapi.tiangolo.com/) framework

- initiate a basic fastapi project structure
- start the building blocks of the routes metrics and inference

Although incomplete, thanks to the self-documentation, there are snippets of documentation
image

What is not included : 
[] CI
[] Security aspects

There is still a lot of work to be done, but perhaps this will help to start the discussions about this api

Feedbacks are welcome


frgfm (changes requested):
I suggest moving everything outside of the `pyroengine` folder similarly to https://github.com/pyronear/pyro-risks ;)

And there are a few unittests to be passed also!

fe51:
I have moved everything in a dedicated folder called `server`, added license headers and dummy tests for the routes !

fe51:
Once again, this PR provides the structure of the local server, content of routes will come with others PRs ! 

@frgfm I would be happy to have your feedback :)

server/src/requirements.txt
```diff
@@ -0,0 +1,7 @@
+fastapi==0.61.1
+uvicorn>=0.11.1
+python-multipart
+
+# dev
+pytest>=5.3.2
+requests
```
frgfm, line 7: I suggesting moving this to a `requirements-dev.txt`

server/src/app/config.py
```diff
@@ -0,0 +1,15 @@
...
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+
+import os
+
+PROJECT_NAME: str = 'Pyronear local system API'
+PROJECT_DESCRIPTION: str = 'API for pyronear systems interactions on sites'
+API_BASE: str = 'api/'
+VERSION: str = "0.1.0a0"
+DEBUG: bool = os.environ.get('DJANGO_DEBUG', '') != 'False'
```
frgfm, line 13: No django over here :)

server/src/app/config.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+
+import os
+
+PROJECT_NAME: str = 'Pyronear local system API'
+PROJECT_DESCRIPTION: str = 'API for pyronear systems interactions on sites'
```
frgfm, line 10: Perhaps Webserver would be more accurate

server/src/app/api/schemas.py
```diff
@@ -0,0 +1,37 @@
...
+    mem_available_GB: float = Field(None, example=1.23)
+    cpu_usage_percent: float = Field(None, example=51.8)
+
+
+class MetricOut(MetricIn):
+    pass
+
+
+class InferenceOut(_CreatedAt, _Id):
+    pass
```
frgfm, line 37: I'm not following? If we don't implement the inference route, let's remove it completely but I doubt the output schema is correct :sweat_smile:

server/src/app/api/routes/metrics.py
```diff
@@ -0,0 +1,24 @@
...
+
+from fastapi import APIRouter
+
+from app.api.schemas import MetricIn, MetricOut
+
+router = APIRouter()
+
+
+@router.post("/", response_model=MetricOut, status_code=201, summary="Send metrics from a device")
+async def send_metrics(payload: MetricIn):
```
frgfm, line 15: Suggestion: `log_metrics`

server/src/app/api/routes/inference.py
```diff
@@ -0,0 +1,36 @@
...
+                    ):
+    """
+    Send img from a device based on the given information in order to predict smoke
+
+    Below, click on "Schema" for more detailed information about arguments
+    or "Example Value" to get a concrete idea of arguments
+    """
+    # some async operation could happen here
+
+    # We might need a backround task here for inference ?
```
frgfm, line 33: Could you elaborate?
Are you implying that we don't need the output? If not, I don't see why background task would help

server/src/tests/routes/test_metrics.py
```diff
@@ -0,0 +1,38 @@
...
+    "mem_available_GB": 1.23,
+    "cpu_usage_percent": 51.8
+}
+
+
+def test_send_metrics_bad_type(test_app):
+
+    response = test_app.post("/metrics/", data=json.dumps(BAD_PAYLOAD))
+
+    assert response.status_code == 422
```
frgfm, line 38: Let's use the `@pytest.mark.parametrize` decorator to refactor this ;)

frgfm (commented):
Thanks for the PR! I left some comments!

server/src/app/api/routes/inference.py
fe51: The client (here the pi camera that send the image) does not have to wait for the inference & alerting step to continue, we can juste return something like  "Accepted" and the process run in background. That is why I think a [backround task](https://fastapi.tiangolo.com/tutorial/background-tasks/#background-tasks) could be relevant but I may be going the wrong way !

server/src/app/api/schemas.py
fe51: See discussion with inference route, as long as inference takes place in background, the output does not matter so much (but you are right, the schema is not correct, and might be better to at least return success/fail of the post (but this info is already in the status_code of the reponse). 

Or maybe, no output data ? thus, not InferenceOut class

server/src/app/config.py
fe51: Pyronear local webserver?

server/src/app/config.py
fe51: Nice catch :)

fe51:
> Thanks for the PR! I left some comments!

I will quickly implement some fixs related to your comments

server/src/tests/routes/test_metrics.py
fe51: Would be cleaner indeed ! Though, do not if my refactor is the cleanest way to use the decorator.

server/src/app/api/schemas.py
frgfm: Alright but then let's not return anything :)

Here you define a schema with two fields : `created_at` and `id`. The `id` field is set to a fixed value (so it's not needed) and the created_at isn't used so the route will return a 422 status code :sweat_smile:

server/src/app/api/routes/inference.py
frgfm: Well it depends but if the central pi does have a camera, we might want to raise an alert right after analyzing (if it detects something). So are we doing this only for cameras --> central interactions?

server/src/requirements.txt
frgfm: But they also need to be removed from this `requirements.txt` :sweat_smile:

server/src/tests/routes/test_metrics.py
```diff
@@ -0,0 +1,32 @@
...
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+
+import json
+import pytest
+
+
+PARTIAL_PAYLOAD = {
+    "id": 1,
+    "created_at": "2021-03-26T16:35:09.656Z",
```
frgfm, line 13: up to you, but we could fill the created_at on the API side and not request it in the payload

frgfm: that way the client doesn't even need an internal clock

frgfm (changes requested):
Thanks for the edits! A few comments left

server/src/requirements.txt
fe51: ahah yes I know (I will remove once we are okay with everything else, it let me change nothing to run the tests locally :D )

server/src/app/api/routes/inference.py
fe51: If we consider a camera on the central pi, it could send image to the local webserver, and it will work like any other cameras. 
But I might have misunderstood something. Concretely, the background task is launch just after the response of the request right ? So right after sending it, it just let the device that send image not to wait during the inference process.

server/src/app/api/schemas.py
fe51: If the inference run as a background task, what should we return to the device client ?

server/src/tests/routes/test_metrics.py
fe51: Yes, I see your point !

As long as the central device is connected to the internet, it should be quite on time and If we were going to only rely on a date created_at on API side, if the device loses  internet connection (without being turned of), when it recovers internet, the datetime of potential alerts in the meantime could be incorrect.

fe51: oups we are talking about metrics routes here. well, we could avoid created_at indeed, can be handled either webserver side or API side (but with the point detailed in the above comment)

server/src/app/api/schemas.py
frgfm: An empty payload since we don't even know if the background task will succeed. We have very limited information to communicate (simply that we attempted launching a background task)

server/src/app/api/routes/inference.py
frgfm: Well yes it runs afterwards but it's asynchronous so the request won't wait for the background task to be executed! It is a viable option sure, depends on what you prefer

server/src/tests/routes/test_metrics.py
frgfm: considering that when the webserver won't be connected to the internet, which means that the device won't either (since the webserver runs on the central pi), I would go for the second one but whichever you prefer :)

fe51: Well, I think we agree, having created_at on API side make things easier. 

We can start with this option. After tests phases, if it occurs that losing internet connection is quite frequent, we could consider other options (in all cases, without internet, not being able to quickly alert that a fire has started make us fail our princpal aim ^^ :) )

server/src/app/api/routes/inference.py
fe51: I would to try this way first and refactor later if needed !

server/src/tests/routes/test_metrics.py
fe51: I have updated consequently, removing created_at !

fe51:
@frgfm this is eventually quite empty web server that is eager to be completed :)

I think this okay now and we can go on with the next PRs to add core features !

fe51:
@frgfm when you find the time to look at it ;)

frgfm (commented):
I think there is only the timestamp missing on the metric (you have to set it with default value in the schema)! And we're good to go!

server/src/app/api/schemas.py
```diff
@@ -23,15 +23,11 @@ class _Id(BaseModel):
...
 
 # Metrics
-class MetricIn(_CreatedAt, _Id):
+class MetricIn(_Id):
     cpu_temperature_C: float = Field(None, example=17.3)
     mem_available_GB: float = Field(None, example=1.23)
     cpu_usage_percent: float = Field(None, example=51.8)
 
 
 class MetricOut(MetricIn):
```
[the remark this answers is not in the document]
fe51: As long as timestamp are handled on API, do we need a timstamp as an input ? (cf our previous discussion about it )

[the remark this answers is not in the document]
frgfm: No, I meant: you HAVE to add them so that the API sets it itself :sweat_smile: 

The other solution is that: whatever the database you will send this to, will need to add its timestamp. If that's OK with you, we can indeed merge.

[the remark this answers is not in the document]
fe51: My bad, that is way I added _created_at in the first place ^^
 
I have fixed it, you can and approve and this PR will be ready to merge :)

frgfm (approved):
Thanks for the edits!

Issue #5, opened 2020-09-22
Add Rapberry Pi tests


MateoLostanlen:
## 🚀 Feature
We need to create a test environment for rapsberry pi using a docker. It is important that these tests are part of the CI

frgfm:
Big issue so far: docker will not be able to create an ARM environment on common architectures :/
I haven't found my way around that so far...

MateoLostanlen:
Damn... I just red your comment, i'm trying for 2 days haha

MateoLostanlen:
@frgfm maybe we can try this : https://dev.to/l04db4l4nc3r/self-hosted-github-actions-using-raspberry-pi-212m

And testing raspberry specific features only for release. What do you think ?

MateoLostanlen:
Just found this, fake raspberry package https://pypi.org/project/fake-rpi/

frgfm:
@MateoLostanlen Regarding self hosting, I thought about this, but apart from having a dedicated Raspberry pi connected to the internet which would run all the CI, I don't see how we could managed it :/

I'll look into the fake raspberry package :ok_hand:

MateoLostanlen:
If it's Work the fake package might be a good solution. If not, we could buy a dedicated pi (maybe a pizero)

PR #16, opened 2021-04-06
Test patch

MateoLostanlen:
Following the discussions in #5 I propose you the following solution inspired by this repo : https://pypi.org/project/fake-rpi/
We create fakes Raspberry pi library classes for the tests, and we use the pi_patch module to replace the real ones by the fake ones during the tests. I find this solution quite elegant and it solves our problem. It's not perfect because we don't reproduce the exact working conditions but it's similar

What do you think ?

### DIFF
--- a/test/pi_patch.py
+++ b/test/pi_patch.py
@@ -1,0 +1,10 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+import sys
+import fake_gpiozero
+
+
+sys.modules['gpiozero'] = fake_gpiozero 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 pyrovision
+psutil
--- a/test/fake_gpiozero.py
+++ b/test/fake_gpiozero.py
@@ -6,6 +6,6 @@
 
 class CPUTemperature:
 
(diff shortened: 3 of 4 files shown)


MateoLostanlen:
The broken link is the one of the docs and flake8 can't be done better in pi_patch module

test/pi_patch.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the GNU Affero General Public License version 3.
+# See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
+
+import sys
+import fake_gpiozero
```
fe51, line 7: Does `fake_gpiozero` should be added to the dev requirements ?

fe51 (commented):
Thanks for the PR ! 

I am not so much knowledgeable, but its seems a nice way to tests code using ARM elts !

Except requirements question,  looks okay to me :)

test/pi_patch.py
MateoLostanlen: no we do not whant to use it in the real package, it's just for testing on github

MateoLostanlen: moreover it's not a package so can't be installed ;)

fe51: ho okay! So how to install it ?

fe51 (approved):

test/pi_patch.py
fe51: okay my bad I just get the tricks ^^Nice

PR #17, opened 2021-04-06
reorganize repo

MateoLostanlen:
This PR replace #13

The purpose of this PR is to reorganize the repo. Now we have:

    a pyronear_predictor class that loads the last pytonear model and makes a prediction on a given image.
    A Pyronear Engine class that manages the whole process of wildfire detection, from the image loading to the api call.

Pyro Engine functioning is currently simplistic, it should be improved in a future PR. Here we launch an alert each time we detect fire (so every n second when a fire outbreak), this is obviously not possible in reality.

the plaform folder have been removed, the api in #15 will init and use the pyronear Engine

We should merge this after #16 to include test patch


fe51:
Hey, thanks for the PR !

Currently simplistic but it will help to reach our MVP !

Plus, I like the idea to have pi_utils and its related requirements. In terms of updating, it's a good way to imagine that the central pi updates the repo and then "copy" some scripts to the camera pi (the pi_utils) with the associated installation requirements

fe51:
Also, do you want to add tests related to scripts in 'pyroengine' folder in a next PR ?

MateoLostanlen:
No I Will add the test here. I'am waiting for api credential to test the call in CI. 
I'v open the PR à bit early, why I did not requested review yet

fe51:
ho Okay indeed ! 
(I dot not have the creds btw :) )

pyroengine/pi_utils/requirements.txt
```diff
@@ -0,0 +1 @@
+gpiozero>=1.5.1
```
fe51, line 1: Does `psutil` package has to be added in this requirements file  ?

fe51: Or more precisely, is this psutil package used elsewhere than in pi_utils in the repo? 
If not, it should be better to require it here and not in the general requirements file 

(Thinking about this point because, some of pi_utils scripts will be used on pi camera, hence, these are some of the requirements for the pi camera)

fe51: well might be the same for `pandas` if not used elsewhere

MateoLostanlen: psutil is already in the package requierements so no need to add it twice. We need it to be installed for testing, this is way i moved it in package requirements. Idem for pandas

fe51: Indeed ! 

Not blocking for the PR, but, would it be okay to still add those libs here ? Would be redondant, but we would exactly know what is needed for picameras .    Just an open question, we might have other solutions for a nice set up :)

fe51 (approved):

PR #19, opened 2021-04-10
Changes in pyro-engine for rpi

chloeskt:
### Hi everyone ! 

This is a small PR which aims at working this repository easy to use once it is installed on our raspberry pis. For that I propose the following changes: 

#### Regarding the root of the repo

- add at the root of the repo, a `main.rpi.requirements.txt` file: this file gives the requirements necessary to install on the main rpi regarding its use of pyro-engine. These are not the same as `requirements.txt` due notably to the fact that pyro-vision cannot be installed that easily (itself due to the fact that we are using specific compiled versions of pytorch and torch-vision`. The pyro-sys-setup repository takes care of installing pyro-vision (as a library) on the main rpi through a specific process. Plus, I believe that we should enforce the requirements on all rpis (i.e. put ==) as even a small change could cause issues (better to be on the safe side). 

#### Regarding the directory `pyroengine/pi_utils`

- rewrite `monitor_pi.py`: 
          - first to make it follow PEP8 guidelines regarding snake cases
          - second to make it communicate to the local webserver and send it the data in the required format
          - right now, the local webserver needs an "id" to be send over on the route "/metrics". I have hard-coded the number 0 for it in `monitor_pi.py` in order to let you decide how you want the id to be generated (either by the webserver itself or by the device when running the script)
- add a template of .env file which is used in `monitor_pi.py` (need to put the IP adress of the main rpi running the local webserver and the corresponding port)
- rename camera.rpi.requirements.txt to make it easier to understand + add strong requirements needed to be installed to the rpi with camera 

#### Update of the tests concerning monitor_pi.py

Feedback is welcomed as usual 🤗 

NB: Once this PR has been merged, I will propose the PR integrating this repository on our devices using `pyro-sys-setup` on the latter repo


MateoLostanlen (approved):
Thanks for this PR ! All good for me

PR #20, opened 2021-04-11
Inference using engine

MateoLostanlen:
The goal of this PR is to update the api inference function by using the pyronear engine to run the prediction

This PR is not ready yet, we are juste sharing progress with @fe51 do not review it :)


server/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     build:
       context: ..
       dockerfile: ./server/src/Dockerfile
-    command: uvicorn server.src.app.main:app --reload --workers 1 --host 0.0.0.0 --port 8000
+    command: uvicorn server.src.app.main:app --reload --workers 1 --host 0.0.0.0 --port 80
```
fe51, line 8: Just for reference, why 80 instead of 8000 ?

pizero/runner.py
```diff
@@ -0,0 +1,23 @@
+import io
+import time
+import picamera
+from PIL import Image
+import requests
+
+
+url = "http://192.168.1.62:8000/file" # api url
```
fe51, line 8: Considering, the change in docker compose, should not it be 80 ?

fe51: So far, okay for the url as it is, but will be in the end in a config file or var env ! (I might came with a suggestion soon, a "fleet-config-file" to refer to (not versionned in the repo), and the url would based on those info)

fe51: Also, does it work as it is ? 

as long as in `server/src/app/main.py` , l19, the inference route is prefixed as inference ` prefix="/inference`, should not it be
url = "http://192.168.1.62:8000/inference/file" ?

fe51 (commented):
Hey ! Thanks for the PR and pleased to see expected features coming up !

I have left some comments and questions !

server/docker-compose.yml
MateoLostanlen: no specific reason, it wasn't on purpose. juste I didn't put back the right port after testing stuff. I will move everything back to 8000

pizero/runner.py
MateoLostanlen: Yes I need to a something to proper setup ip adress. About the port yes this is an error ;)

fe51: Okay :)
Looks good to me then 👍

fe51 (approved):
Looks good to me now !
(Considering a proper set up for url etc will be implemented soon)

fe51:
After fixing the linter :)

PR #21, opened 2021-04-11
fix: Forced frame to RGB mode

MateoLostanlen:
Quick fix to avoid issue with PNG image with alpha channel


frgfm (approved):
thanks for the fix!

fe51 (approved):

PR #24, opened 2021-04-11
Allow webserver pyroengine use

fe51:
This small PR aims at allowing webser to access pyroengine 

Plus, the restart option has been added in order to handle webserver reboot.

I have some issues with pyroengine dependencies (will be fixed in another PR) so `server/src/requirements.txt` has not be updated. 

Feedback are welcome !


chloeskt (approved):
Hi @fe51, thanks for this PR ! Looks good to me

PR #25, opened 2021-04-12
Solve vision dependencies compliance - alpine to slim docker image 

fe51:
This PR aims at successfully building the webserver with pyro-vision dependencies. 

Some builds of pyro-vision requirements seems to not be available for alpine docker image ([open-cv]( https://stackoverflow.com/questions/50950588/trouble-installing-opencv-in-docker-container-using-pip), torch), and I cannot succeed to build the image with `pyro-vision` in requirements. 

Hence, this PR provides a **Dockerfile based on slim image** that allow webserver to be built with pyro-vision package

Btw : Some [resources](https://pythonspeed.com/articles/base-image-python-docker-images/) states that python could be faster on slim image than on alpine, that is a good point too. 

As long I am not so much knowledgeable, maybe some commands in the dockerfile could be removed to lighten the image. (https://github.com/docker-library/python/blob/master/Dockerfile-slim.template)

So far, with this set up, it is working. I am not so confident that it will work on pi with ARM processor, but this point should addressed in another issue/PR.

Feedbacks are welcome !

### DIFF
--- a/server/src/Dockerfile
+++ b/server/src/Dockerfile
@@ -19,9 +19,9 @@ ENV PATH /usr/local/bin:$PATH
 ENV LANG C.UTF-8
 
 # runtime dependencies
-RUN set -eux; \
-	apt-get update; \
-	apt-get install -y --no-install-recommends \
+RUN set -eux \
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
 		ca-certificates \
 		netbase \
 	&& rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


MateoLostanlen (approved):
Thanks for this PR. I accept this PR to be able to continue the tests but I will open a new one soon to load an image already built from dockerhub with torchvision and pyrovision for a faster deployment

fe51:
Sounds great to hear! Especially since the internet speed on site will not be very fast !

PR #30, opened 2021-04-15
chore: Add workflow to publish docker image

MateoLostanlen:
The goal of this PR is to add a workflow to publish pyro-engine docker image on docker hub for each release


.github/workflows/publish-docker-image.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
+          username: pyronear
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
frgfm, line 16: I only suggest renaming it DOCKERHUB_PW or DOCKERHUB_PASSWORD (there are several docker registries, one on github by the way and we may, in the future, push the image to multiple repositories)

frgfm (commented):
Thanks for the PR! Only a small comment on the secret naming but the rest looks quite alright!

.github/workflows/publish-docker-image.yml
```diff
@@ -0,0 +1,18 @@
...
+  push_to_registry:
+    name: Push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
+          username: pyronear
```
frgfm, line 15: On second thought, let's put this as a secret as well since our repos are open source :pray:

.github/workflows/publish-docker-image.yml
MateoLostanlen: Well spotted, thanks !

.github/workflows/publish-docker-image.yml
MateoLostanlen: I hesitated about this but the info is not secret, right?

frgfm: half true, imagine that for one of our service, there is no way to have org credentials: one of us will have to add her/his personal creds. I agree that the risk is low, but it easily decreases the risk from having a bot figuring out one of our two credentials for a service :sweat_smile:

ThibaultDac (approved):
Approved
Best Practices:  Don't write the username (not important here because it's public)

frgfm (approved):
Thanks!

.github/workflows/publish-docker-image.yml
MateoLostanlen: yes sur, it cost nothing at the end :)

PR #31, opened 2021-04-17
Add security

fe51:
This PR aims at adding security to the webserver with OAuth2 specification in order to make it safe (only the fleet of pi camera and central pi will be able to use routes)

I has been implemented with the help of the dedicated [Fastapi documentation](https://fastapi.tiangolo.com/tutorial/security/).

### Features

Below are the main steps :

- Add security using oauth2 with jwt tokens and password hashing
- Add login route to retrieve token
- Add auth requirements to inference and metrics routes
- Updates unit test using [dependency_overrides](https://fastapi.tiangolo.com/advanced/testing-dependencies/). (test to properly check security flow has not been implemented, maybe to add later )
- Add new schemas related to not new workflow, and updates metricIn removing Id as long as device will be detected by auth

### Usage

- To date, registered devices/users are stored in a json file (fleet_accesses) which (a bit as an env file)

The file looks like:

```javascript
{
    "camera_1": {
        "username": "camera_1",
        "hashed_password": "$2b$12$EixZaYVK1fsbw1ZfbX3OXePaWxn96p36WQoeG6Lruj3vjPGga31lW",
        "disabled": false
    },
    "michel": {
        "username": "michel",
        "hashed_password": "$2b$12$EixZaYVK1fsbw1ZfbX3OXePaWxn96p36WQoeG6Lruj3vjPGga31lW",
        "disabled": false
    }
}
```

(to generate the hash of the password, us get_password_hash from security.py  )

- Hence, to use the webserver with ease, a pi camera/device need to retrieve its token using login route and then add it to headers of routes that require auth (all others)

Any feedback is welcome :)


frgfm:
Thanks a lot for the PR!
Generally speaking, I'm not getting why we take the current active user?
Considering the usage of this repo and this webserver, we are only dealing with devices :) I think the PR should be updated accordingly before moving on to review

fe51:
Indeed, users of the webserver, are here, devices. I will change typo accordingly to make it clearer!

server/src/tests/routes/test_inference.py
```diff
@@ -7,10 +7,12 @@
 import requests
 
 
-def test_inference(test_app):
+def test_inference(test_app, mocker):
+
+    mocker.patch('app.api.deps.get_current_active_user', return_value="kikou")
```
frgfm, line 12: needs to be updated

server/src/app/config.py
```diff
@@ -4,8 +4,23 @@
...
 LOGO_URL: str = "https://github.com/pyronear/PyroNear/raw/master/docs/source/_static/img/pyronear-logo-dark.png"
+
+
+# security
+JWT_ENCODING_ALGORITHM = "HS256"
+ACCESS_TOKEN_EXPIRE_MINUTES: int = 60 * 24 * 365 * 10
+
+# to get a string like this run:
+# openssl rand -hex 32
+SECRET_KEY = "09d25e094faa6ca2556c818166b7a9563b93f7099f6f0f4caa6cf63b88e8d3e7"
```
frgfm, line 22: Time to get lectured :sweat_smile: 
**We NEVER EVER EVER EVER EVER put a secret key in a commit / PR / etc**. Even if you delete it in your next commit, the PR history will keep it.

So :
- from a security perspective, let's never do this again
- this secret key is compromised, so we will need to generate another.
- Github Secrets and `.env` files are your friends ;)

frgfm (changes requested):
Thanks for the edits!

However, there is a big security issue to handle (cf my comment)
Besides this, I think it would be better to open an issue and discuss features & implementations. I'm starting to think that we are doing pyro-api all over again for something that only needs to support 3 interactions :sweat_smile:

server/src/app/config.py
```diff
@@ -4,8 +4,23 @@
...
 VERSION: str = "0.1.0a0"
 LOGO_URL: str = "https://github.com/pyronear/PyroNear/raw/master/docs/source/_static/img/pyronear-logo-dark.png"
+
+
+# security
+JWT_ENCODING_ALGORITHM = "HS256"
+ACCESS_TOKEN_EXPIRE_MINUTES: int = 60 * 24 * 365 * 10
+
+# to get a string like this run:
+# openssl rand -hex 32
```
frgfm, line 21: Let's not scream out to the world how to brute force our secret key :upside_down_face: 
We'll need to change how it gets generated as well then

server/src/tests/routes/test_inference.py
fe51: Nice catch

server/src/app/config.py
fe51: As long this key is the one from the fastapi tutorial it was not planned to keep this one anyway ! 
Thought it would be easier to test/debug but should have had a debug key mode  and put aBother key in secrets Shard with pyro users

server/src/app/config.py
fe51: Okay ! I have another idea !

fe51:
Hello @frgfm thanks for the feedbacks ! 
Isn’t just a webserver based on fastapi handling security using jwt token ? Looks like pyro-api indeed, a similar architecture male it easier to maintain ? 
also, I agree with issues before implementations but first it has been discussed orally, and secondly seeing the success of others issue I directly dive into implementation. 
However I would be happy to discuss the suitable solution for this webserver .

server/src/tests/routes/test_inference.py
fe51: fixed

fe51 (commented):
Secret key has been removed, and the secret key will be loaded from env file.
Will see soon to handle it via Github Secrets

MateoLostanlen:
> 
> 
> Secret key has been removed, and the secret key will be loaded from env file.
> Will see soon to handle it via Github Secrets

@fe51 I can add the github secret if you need, contact me via slack :)

frgfm (commented):
Since the PR is a bit massive and I can't dive deep enough to confidently approve, now that you have set unittests, could you add a github workflow job to run them?

It would be easier for review :pray:

MateoLostanlen:
All good, thanks :)

MateoLostanlen (approved):

PR #33, opened 2021-04-21
create env file and remove not used import

chloeskt:
Hi everyone ! 

This short PR is made in order to remove the hardcoded version of the url in the `pizero/runner.py` script. It is better to put the IP and PORT in a separate environment file under the model of `python.env.dist`. Also I removed an import which was not used in this script, namely `PIL` (if needed after for an improvement of the script, it could be added at this time).

Consequently I updated the `README.md` and the previous env file in `pi_utils` so that all variables have the same name.

Also I have a more general question: why was this folder `pizero` created ? I believe that the `runner` script could be put inside `pi_utils` since this directory already has pizero's requirements + the monitor script + the environment file corresponding to the IP and PORT of the local webserver which will run on the main rpi. I don't see why we would need to create separate directories for these two scripts, in my opinion they go hand in hand. What was the motivation ? 

As always, feedback is welcomed 🤗


fe51 (approved):
Thanks for the PR ! Looks good to me and I agree with your interrogation ! Let's discuss you question about pizero folder before merging :)

chloeskt:
Yes let's wait for @MateoLostanlen's input ! :)

MateoLostanlen:
@chloeskt agree to move the pizero runner and delete the folder 😉

MateoLostanlen (approved):

chloeskt:
Alright ! In that case, this evening I shall modify this in this current PR then and ask for your reviews again. Or do you prefer that I open a new PR ?

MateoLostanlen:
1 PR = 1 action :)

MateoLostanlen:
I can open the new PR if you want

chloeskt:
> I can open the new PR if you want

I already did everything, I just want to check that all requirements, the flow, etc is working on rpi

PR #34, opened 2021-04-21
Refacto pizero directory into file in pi_utils and update consequently

chloeskt:
Hello everyone,

Following discussion in #33, this PR aims at:
- removing directory `pi_zero` to integrate it in `pi_utils`
- consequently update `camera.rpi.requirements.txt` and `README.md`
- keep only the names `WEBSERVER_IP` and `WEBSERVER_PORT` throughout all the scripts for clarity 

There is the issue that `runner.py` has no tests which reduces heavily the coverage but I guess that this was already the case when this script was merged into master. One should add tests later I guess.


fe51 (approved):
Hello ! Thanks for the quick PR !

PR #36, opened 2021-04-26
Rewrite of runner.py to make it similar to monitor_pi.py

chloeskt:
Hi everyone ! 

I have taken the liberty of rewriting `runner.py` so that it matches the structure of `monitor_pi.py`, namely:
- create a class
- add logger (because I think that in the future it might be nice to catch the errors if some happens corresponding to RequestException, but for that I shall create rotating logs first, but anyways, not a priority)
- add corresponding unit tests

I proposed this new structure, any feedback is welcomed ! 🤗 

However I have a question regarding the code that was already in `runner.py`. \
The instantiation `stream = io.BytesIO()` was done two times: one before the loop (line 22) and one in the loop (line 27). Why is that ? \
The one in the loop overtakes the previous one. I'm not familiar with `io.Bytes`: do we need to re-instantiate it everytime we want to capture a stream ? I have supposed that it is the case in the code that I propose here. However if it's not, then we could move `stream = io.BytesIO()` in the `__init__` I guess. What do you think ? 

Also disclaimer: I could not test the entire code on my rpi since the dockerfile is not working, the `send_stream` method has thus not been tested, only the `capture_stream` and `run` ones have been. 

Capture d’écran 2021-04-26 à 22 43 26

### DIFF
--- a/pyroengine/pi_utils/runner.py
+++ b/pyroengine/pi_utils/runner.py
@@ -3,13 +3,14 @@
 # This program is licensed under the GNU Affero General Public License version 3.
 # See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
 
-
+import logging
 import io
 import os
 import time
 import picamera
 import requests
 from dotenv import load_dotenv
+from requests import RequestException
 
 load_dotenv()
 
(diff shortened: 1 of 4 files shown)


fe51 (approved):
OKay to me, let's wait from @MateoLostanlen  answer to be sure :)

MateoLostanlen:
Hello @chloeskt, yes we need to re-instantiate or at least to free memory. I need to check if there is a better solution for that. 

It seems that `stream.seek(0)` could work, I need to test it :) 

I think you can leave as it is for now :)

MateoLostanlen (approved):

PR #39, opened 2021-04-29
Take picture in pi camera, send and write in pi main

blenzi:
This PR aims at adding a way to take a single picture in a pi camera, send it to a main RPI and write it locally. This will simplify the hardware installation process and possibly future checks.

This is just a draft code / minimal example to initiate the discussion, no checks and tests added. If this is included, the next step could be to have an interface to request pictures from specific pi camera devices and view the latest ones in the local pi main web server.

### DIFF
--- a/pyroengine/pi_utils/runner.py
+++ b/pyroengine/pi_utils/runner.py
@@ -2,7 +2,7 @@
 
 # This program is licensed under the GNU Affero General Public License version 3.
 # See LICENSE or go to <https://www.gnu.org/licenses/agpl-3.0.txt> for full license details.
-
+import argparse
 import logging
 import io
 import os
@@ -41,10 +41,10 @@ class Runner:
         stream.seek(0)
         return {"file": stream}
 
-    def send_stream(self, files):
+    def send_stream(self, files, **kwargs):
         try:
             response = requests.post(
-                self.webserver_url, files=files
+                self.webserver_url, files=files, **kwargs
             )  # send image to local webserver
             response.raise_for_status()
         except RequestException as e:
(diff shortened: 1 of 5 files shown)


chloeskt:
Hi @blenzi, thank you very much for this 🤗 

I am not convinced that we should edit `runner.py` and the route `inference` in such a way for two main reasons:

- in programming I like the paradigm **"one class" = "one action"**, namely here capture a stream (i.e. bytes) and send them to webserver to inference. You might argue that there are two actions here, and yes it is true. However they have an internal coherence. I am afraid that adding what you have done so far breaks that coherence. 
        - first, you want to capture a stream and then write it 
        - second, it is not for the inference but only for us to check. Hence you are also modifying the route `inference` except it should not be the case. Again I believe that **"one route" = "one action".** Otherwise it is starting to be messy.
  
- you are adding complexity to a class and to a route where I believe that it would be more efficient to separate into parts that have internal coherence. `Inference` and `runner` where already doing what they were supposed to. **By adding complexity here, you are adding possibilities of errors and bugs which would make the debugging on the rpi for instance harder.** By writing classes or functions that have internal coherence and only a specific role, we make it easier for us to debug later.

**Your idea is nice but might be too complicated the way you initiate it and prone to bugs in my opinion**. Might be easier if you create a separate class doing exactly what you want it to do and create the corresponding route in the webserver. (Also it will make testing through unit tests and on your rpi easier !)

What do you think ?

MateoLostanlen:
Hi @blenzi , thanks for the PR, I totally agree with @chloeskt . This feature is a real plus but it doesn't belong in the runner simply because we won't use it while our application is running.

blenzi:
Hi @chloeskt , @MateoLostanlen , thanks a lot for the comments. 

That was just a way to add minimal code changes and flexibility. I am happy to separate the classes and routes. However, to avoid duplication (or inheritance) and follow the *one class = one action* logic, maybe it would be easier to have a separate class just to take pictures, that could be used by the one that sends them for inference and a new one that sends them for writing, each one using one route. What do you think?

MateoLostanlen:
@blenzi, sorry for the delay and thanks again for this. It seems good to me. 
We really need to improove code cove, but I guess it will lead to code duplication here. 
Let's deal with this in another PR

MateoLostanlen (approved):
Hi @blenzi, sorry for the delay and thanks again for this. It seems ok for me !
We realy need to improove codecove, but without code duplication it implies reoganising a bit the tests in this repo. Let's see this in another PR !

MateoLostanlen:
Can you just merge master beforer merging ?

PR #40, opened 2021-04-30
Add bash script for internet failures scenario

chloeskt:
Hi everyone, 

In this PR you can find a proposition for internet failures scenario. This script bash will then be added on the main rpi (using pyro-sys-setup) and will be put in a cron job so that in it run every 15 minutes for instance (the number here is chosen arbitrarily, if you have better ideas, feel free to tell me).

The flow is the following:
```
Ping internet: 
	- if success: 
		Ping vpn:
		- if success: ok exit 
		- if failure: restart vpn

	- if failure: restart of network interfaces (all interfaces because I don't know which internet the 4G modem will take)
sleep 60 sec
Ping internet again (to make sure that we have internet):
	- if success: ok exit
	- if failure: reboot
```

What do you think ?

### DIFF
--- a/pyroengine/pi_utils/check_internet_connection.sh
+++ b/pyroengine/pi_utils/check_internet_connection.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# This script performs:
+# Ping google.com:
+#	- if success:
+#		Ping internal vpn ip:
+#		- if success: ok exit
+#		- if failure: restart vpn
+#
+#	- if failure: restart of network interfaces
+# Sleep 60 sec
+# Ping google again (to make sure that we have internet):
+#	- if success: ok exit
+#	- if failure: reboot
+
+set -u
+
+PING_CMD=('ping' '-c' '1' '-W' '10' 'google.com')
+
+if "${PING_CMD[@]}";
+then
+    iplocalhostvpn=$(ifconfig tun0 | awk '/inet / {print $2}')
+    PING_CMD_VPN=('ping' '-c' '1' '-W' '10' "$iplocalhostvpn")
+    if "${PING_CMD_VPN[@]}";
+    then
+          exit 0;
+    fi;
+
+    sudo systemctl restart openvpn@client;
+fi;
+
+sudo service networking restart;
+sleep 60;
+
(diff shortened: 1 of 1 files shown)


blenzi:
Thanks @chloeskt, looks good to me! I guess there is no easy way to add a test? One suggestion is to call it `check_internet_connection` and add the description above to the file.

MateoLostanlen:
Hi @chloeskt , thanks for the PR, Just one question. If there is a problem, your script will reboot the pizero. What happens after that? How do you restart the runner? Shouldn't this reboot be managed by the pi central to restart the pizero afterwards ?

chloeskt:
> Hi @chloeskt , thanks for the PR, Just one question. If there is a problem, your script will reboot the pizero. What happens after that? How do you restart the runner? Shouldn't this reboot be managed by the pi central to restart the pizero afterwards ?

Hello Mateo ! This script bash will only be run on the main rpis since they are the ones with the vpn and the internet connexion not the rpi with camera. If the main rpi reboots, the monitor script will be launch again as it is in a cron job starting on reboot. The rpi with camera are not affected by this bash script.

chloeskt:
> Thanks @chloeskt, looks good to me! I guess there is no easy way to add a test? One suggestion is to call it `check_internet_connection` and add the description above to the file.

Indeed no way to test on github. I've tested it a couple of times on my rpi of course, seemed to work fine. Also I integrated your suggestions, thanks !

MateoLostanlen:
> 
> 
> > Hi @chloeskt , thanks for the PR, Just one question. If there is a problem, your script will reboot the pizero. What happens after that? How do you restart the runner? Shouldn't this reboot be managed by the pi central to restart the pizero afterwards ?
> 
> Hello Mateo ! This script bash will only be run on the main rpis since they are the ones with the vpn and the internet connection not the rpi with camera. If the main rpi reboots, the monitor script will be launch again as it is in a cron job starting on reboot. The rpi with camera are not affected by this bash script.

ok sorry I misunderstood. It's perfect then.

MateoLostanlen (approved):
All good, thanks !

PR #41, opened 2021-05-05
v0 of update flow script

chloeskt:
Hi everyone, 

In light of this [PR on sys-setup](https://github.com/pyronear/pyro-sys-setup/pull/18/files), I propose the following Python Class to help launch the updates. The flow is the following: 

- get the hash the main rpi should have by contacting the pyro api (still needs to be implemented and I probably won’t have the time)
- compute the difference between this hash (which would be the hash of the last commit on master) and the one that I have stored in "/home/pi/pyroengine_version.txt" (which is the hash of the last commit of the version the main rpi has on it version of pyroengine)
- if different, `self.update` becomes True and the method `self.run` will run. Otherwise, do nothing.

For the `self.run`:
- maximum of 3 trials for each playbook (this is because I believe that if it fails too much, it means that we have an issue somewhere and we need to fix it manually)
- first, update the main rpi (needed to be in first because we have to DL the new version of pyroengine on the main, so that it can update its children). Upon success, launch update for the rpi with camera. 
- If the update of the main has failed (i.e. if the command running ansible returns a non 0 code), the update of the rpi with camera is not launched and logs are stored in « {self.ansible_directory}/{logs_file_name} » (for instance, "/home/pi/ansible_main/logs_update_myself.txt »). 
- Also if the update of rpi with camera has failed, logs are stored in "/home/pi/ansible_main/logs_update_rpi_camera.txt »

So we will have logs on the rpi and also on Slack, we will receive a short recap of the play of the playbooks and we will know if it has failed or succeeded (this is the [documentation](https://docs.ansible.com/ansible/latest/collections/community/general/slack_callback.html)). 

`self.compute_difference` only checks if the two strings are equal. The hash should be the hash of the last commit on master (which is directly retrieved by Ansible with module git (using for instance "{{ pyroengine_output.after }} »).

To see what is done during the updates, please refer to [this PR](https://github.com/pyronear/pyro-sys-setup/pull/18/files).

This is only a v0 I guess, but I prefer to share it with you now.

### DIFF
--- a/pyroengine/pi_utils/updater.py
+++ b/pyroengine/pi_utils/updater.py
@@ -1,0 +1,100 @@
+import subprocess
+
+
+class Updater:
+    """Class to update the software on the raspberries."""
+
+    def __init__(self, pyroapi_url, pyroengine_version_path, ansible_directory):
+        """
+        Params:
+        - pyroapi_url: str
+            Url of the pyronear api
+        - pyroengine_version_path: str
+            Path to the text file which contains the hash of the current version of pyro-engine
+        - ansible_directory: str
+            Path to the directory containing ansible related files on the main rpi (/home/pi/ansible_main)
+        """
+        self.pyroapi_url = pyroapi_url
+        self.pyroengine_version_path = pyroengine_version_path
+        self.ansible_directory = ansible_directory
+        self.update = False
+
+    def get_new_hash(self):
+        """
+        Ask the pyronear api for the hash the device should have.
+
+        Return:
+            str: the new hash, provided by the pyronear api.
+        """
+        raise NotImplementedError
+
+    def compute_difference(self, new_hash):
+        """
+        Compute the difference between the new_hash given by the pyronear api and the
+        current version of the hash stored in self.file_path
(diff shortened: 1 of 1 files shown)


MateoLostanlen:
Hello @chloeskt, I did not follow much the advances on pyro-sys-setup, this v0 is already functional or it is a simple darft ?

chloeskt:
> Hello @chloeskt, I did not follow much the advances on pyro-sys-setup, this v0 is already functional or it is a simple darft ?

Hi I believe that everything is ready on the pyro-sys-setup side. However here one needs to `self.get_new_hash()` for instance. Also one might discuss the proposals that I have done here. This was the purpose of this PR. I won't be able to propose more features before the deployment in Ardeche hence I share with you what I have already done.

MateoLostanlen:
@chloeskt ok thanks, I will try to have a look on this

MateoLostanlen:
This PR is open for too long I close it for the moment

frgfm:
Can we delete the branch as well @MateoLostanlen?

Issue #48, opened 2021-06-06
Codecov fail


MateoLostanlen:
## 🐛 Bug

Codecov doesn't seem to be very stable. We already made a similar change in the [past](https://github.com/pyronear/pyro-vision/pull/83/files), after v1.0.6 now v1 doesn't work anymore (#47). Changing the version to v1.0.5 [seems ](#49) to solve the problem

@frgfm what do you thiink ?

frgfm:
Can you edit & correct the hyperlink you wanted to put in "seems"? The URL brings back to this very page, so I can't answer your question :sweat_smile:

MateoLostanlen:
sorry for that, it was suppose to bring you to #49, but at the end It's not working :/

MateoLostanlen:
Ok the bug comes from me, I push from my pro pc on which there is another github account, my commit was not marked verified so codcove refuses authentication

PR #49, opened 2021-06-06
fix codecov issue

MateoLostanlen:
Fix codecov issue by chanching version, as explained in #48


frgfm:
I doubt this is the solution or it would have worked on this very PR :thinking:

MateoLostanlen:
oh, the test passed but it still not working, I was fooled by the validated test... I will check this

PR #50, opened 2021-06-06
Update engine

MateoLostanlen:
The purpose of this RP is to update the engine.
In this new version, the following features have been added:

- Management of N devices simultaneously, the N pi-zero connected to the main pi
- possibility to modify the detection threshold, 0.5 is not always the best choice
- Management of alert by counter to avoid launching an alert with a false positive. Each time the system predicts a fire the counter increases by 1, on the other hand it decreases by 1 if no fire is detected. When the counter reaches 3, the first alert is launched and then all the following alerts until the counter drops back to zero.
- Media upload system for our database, the engine sends for each device an image over N to the api.


pyroengine/engine/engine.py
```diff
@@ -13,44 +13,127 @@ class PyronearEngine:
...
+    def send_alert(self, pi_zero_id):
+        """Send alert"""
+        print("Send alert !")
         # Create a media
-        media_id = self.api_client.create_media_from_device().json()["id"]
+        media_id = self.api_client[pi_zero_id].create_media_from_device().json()["id"]
         # Create an alert linked to the media and the event
-        self.api_client.send_alert_from_device(lat=9, lon=9, event_id=self.event_id, media_id=media_id)
-        self.api_client.upload_media(media_id=media_id, image_data=self.image.getvalue())
+        self.api_client[pi_zero_id].send_alert_from_device(lat=9, lon=9, media_id=media_id)
```
Akilditu, line 130: `lat`, `lon` will be retrieved by the device API `lat` `lon` rather than config data ? (as in [here](http://pyronear-api.herokuapp.com/redoc#operation/register_my_device_devices_register_post))

Akilditu (commented):
Thanks for the PR ! 

I'm not relevant for the prediction part except few comments seems OK to me on the alerts side of things.

pyroengine/engine/engine.py
```diff
@@ -13,44 +13,127 @@ class PyronearEngine:
...
+        self.api_client[pi_zero_id].upload_media(media_id=media_id, media_data=self.stream.getvalue())
+
+    def save_frame(self, pi_zero_id):
+        """Save frame"""
+        print("Upload media for dataset")
+        # Create a media
+        media_id = self.api_client[pi_zero_id].create_media_from_device().json()["id"]
```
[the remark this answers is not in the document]
MateoLostanlen: beacause I need a media id to send the media to the api with `self.api_client[pi_zero_id].upload_media(media_id=media_id, media_data=self.stream.getvalue())`

I'm not sur to understand

pyroengine/engine/engine.py
MateoLostanlen: done !

Akilditu: Thanks, could you please also add `azimuth`

pyroengine/engine/engine.py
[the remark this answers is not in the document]
Akilditu: I meant why creating a media within `save_frame()` the purpose of sending and alert with media_id seems to be handled by `send_alert()` ?

Maybe I m missing smth here :)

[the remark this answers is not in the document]
MateoLostanlen: we want to upload images regularly even when there is no fire detected. The goal is to enlarge our dataset and to have an idea of the variations that the camera observes

pyroengine/engine/engine.py
MateoLostanlen: according to @frgfm azimuth should be fix for each device, therefore it's not managed here

Akilditu: Same as `lat` `lon` if I m correct ? I do not see the point of letting this empty as send_alert_from_device() does take it as param ...

MateoLostanlen: yes, for me they should not be here either but I am obliged to pass them because the api is designed like that, I did not really understand the logic. @frgfm maybe you can enlighten us

Akilditu: Originally we were suppose to send the actual `lat` `lon` of the fire itself that is why I guess this was meant to be passed with `send_alert_from_device()`.

Akilditu (approved):
Seems good to me !

Akilditu (approved):
Everything's fine !

PR #54, opened 2021-06-14
update docker

MateoLostanlen:
I just realized that the docker engine image does not have the latest version of the api, in my opinion the easiest way to keep the api up to date is to install it via github directly


server/src/Dockerfile
```diff
@@ -44,4 +44,6 @@ RUN set -ex \
     && rm -rf /root/.cache/pip
 
 # copy project
-COPY . /usr/src/app/
\ No newline at end of file
+COPY . /usr/src/app/
+RUN git clone https://github.com/pyronear/pyro-api.git
+RUN pip install -e pyro-api/client/.
```
Akilditu, line 49: why not adding something like this `RUN pip install -r /usr/src/app/requirements.txt \` instead of those lines ?

MateoLostanlen: I've made the change and tested it, it works. You're right, it does look cleaner.

Akilditu (approved):
Looks good to me know, thanks for the PR!

PR #55, opened 2021-06-20
Map hostnames

MateoLostanlen:
We needed to identify the device that sends the image to the central pi. We assigned a hostname to each device. It was therefore necessary to do a mapping between the ip that we retrieve with the image and this hostname. This mapping does not work inside the docker. My proposal is to mount a volume with the docker and to put the mapping json in this volume. Then you just have to launch a cron job with the scipt ip_hostname_mapping.py to update this mapping regularly. The web server inside the docker will then have access via the volume to an updated mapping


server/ip_hostname_mapping.py
```diff
@@ -0,0 +1,21 @@
+import json
+import socket
```
fe51, line 2: Shouldn't we add `socket` to the docker requirement ?

server/src/app/api/routes/inference.py
```diff
@@ -46,8 +47,13 @@ def predict_and_alert(file, ip):
     # Load Image
     image = Image.open(io.BytesIO(file))
 
+    # Get hosname
```
fe51, line 50: Just typo
```suggestion
    # Get hostname
```

server/ip_hostname_mapping.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+with open('data/ip_hostname_mapping.json') as f:
+    ip_hostname_mapping = json.load(f)
+
+inv_map = {v: k for k, v in ip_hostname_mapping.items()}
+ip_hostname_mapping = {}
+hostnames = inv_map.keys()
```
fe51, line 10: style : maybe some blank lines here ?

fe51 (commented):
Hi there ! Thanks for the PR ! 

Just some style and requirement comment. 

I have tested the `ip_hostname_mapping.py`.

Can you state yo have tested on the docker running on pi ?

server/ip_hostname_mapping.py
MateoLostanlen: This should be lunch by a cron jon outside docker. Moreover socket is pre-installed :)

server/src/app/api/routes/inference.py
MateoLostanlen: thanks

server/ip_hostname_mapping.py
MateoLostanlen: ok :)

MateoLostanlen:
> 
> 
> Hi there ! Thanks for the PR !
> 
> Just some style and requirement comment.
> 
> I have tested the `ip_hostname_mapping.py`.
> 
> Can you state yo have tested on the docker running on pi ?

Hi @fe51, thanks for the review, yes it has been tested on my test tower !

server/ip_hostname_mapping.py
fe51: Indeed ! Double indeed :)

However, if this job is not used by the wodker/webserver, shoulnd't it be in pyroengine/pi_utils ?

MateoLostanlen: ok, I've made the change :)

fe51: Perfect, thanks a lot :)

fe51 (approved):

PR #56, opened 2021-06-28
add check connection script

MateoLostanlen:
This very simple script has for goal to ping the main pi from the pizeros to check the connection then to reboot the pizero if this connection is lost. I will create a cron that run this script every 10mn


chloeskt (approved):
Hi @MateoLostanlen, thank you, it seems good to me ! :)

PR #57, opened 2021-07-15
day alert only

MateoLostanlen:
Our network is not trained to work at night, moreover we have false alerts at daybreak, so I propose to send alerts only between 8am and 9pm

### DIFF
--- a/pyroengine/engine/engine.py
+++ b/pyroengine/engine/engine.py
@@ -93,7 +93,7 @@ class PyronearEngine:
                 # Send alert to the api
                 now = datetime.now()
                 current_hour = int(now.strftime("%H"))
-                if current_hour > 8  and current_hour < 21: #Only during the day
+                if current_hour > 8 and current_hour < 21:  # Only during the day
                     self.send_alert(pi_zero_id)
                 self.stream.seek(0)  # "Rewind" the stream to the beginning so we can read its content
 
--- a/pyroengine/engine/engine.py
+++ b/pyroengine/engine/engine.py
@@ -6,6 +6,7 @@
 from .predictor import PyronearPredictor
 from pyroclient import client
 import io
+from datetime import datetime
 
 
 class PyronearEngine:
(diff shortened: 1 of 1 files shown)


pyroengine/engine/engine.py
```diff
@@ -90,7 +91,10 @@ def predict(self, frame, pi_zero_id=None):
             if self.use_api and self.event_appening[pi_zero_id]:
                 frame.save(self.stream, format='JPEG')
                 # Send alert to the api
-                self.send_alert(pi_zero_id)
+                now = datetime.now()
+                current_hour = int(now.strftime("%H"))
```
florianriche, line 95:
```suggestion
                current_hour = d.hour
```

florianriche (approved):
I suggested a change but ok otherwise

frgfm:
> Our network is not trained to work at night, moreover we have false alerts at daybreak, so I propose to send alerts only between 8am and 9pm

Thanks for the PR! 
Apart from the code itself:
- we may need a wider time range. Summertime daylight differs a lot from wintertime :sweat_smile: 
- this needs to be set as local time, but the code seems to be comparing UTC timestamps?
- if we don't send alerts during that time, perhaps it might be wise to keep saving pictures but not analyzing them to ease the load of the raspberry PI?

Akilditu:
> > Our network is not trained to work at night, moreover we have false alerts at daybreak, so I propose to send alerts only between 8am and 9pm
> 
> Thanks for the PR!
> Apart from the code itself:
> 
> * we may need a wider time range. Summertime daylight differs a lot from wintertime 😅
> * this needs to be set as local time, but the code seems to be comparing UTC timestamps?
> * if we don't send alerts during that time, perhaps it might be wise to keep saving pictures but not analyzing them to ease the load of the raspberry PI?

Yes agreed on those points as discussed by @pechouc maybe we should consider doing this on the platform side (filtering alerts displayed on an API such as https://sunrise-sunset.org/api) what do you guys think ?

frgfm:
> Yes agreed on those points as discussed by @pechouc maybe we should consider doing this on the platform side (filtering alerts displayed thanks to an API such as https://sunrise-sunset.org/api) what do you guys think ?

Yes I didn't think about it first, but it would be actually much simpler indeed :+1:

PR #58, opened 2021-07-22
refactor: Switched console print to logging.info

frgfm:
- reordered imports
- changed console print to `logging.info` for a cleaner history of events on the device
- fixed a comment that was added as a bare string

Any feedback is welcome!


fe51 (approved):
Hi FG ! Thanks for the PR !

PR #59, opened 2021-07-22
feat: Added heartbeat at each frame prediction

frgfm:
This PR simply adds a `heartbeat` (ping) at each frame prediction when the frame sender is known.

Any feedback is welcome!


Akilditu (approved):
Thanks for the PR ! 

This will allow us to report on heartbeats logs and build a dashboard to monitor devices activity. 

Quick note : if there is no more heartbeat on one or few pizeros linked to the same tower it means that the issue may be pizero-related. If there is no signal for all pizero on a same tower it mean that the issue si more likely related to the main rpi / 4G hotspot.

PR #60, opened 2021-09-12
chore: Updated requirements.txt

frgfm:
- removes unused `pandas` dependency since it's quite heavy for low-power devices
- unpinned `pyroclient` from dependencies of the main Raspberry (@chloeskt could you review this carefully please? I'm not sure why it was pinned in the first place, so I might have missed something)

Any feedback is welcome!

### DIFF
--- a/main.rpi.requirements.txt
+++ b/main.rpi.requirements.txt
@@ -1,4 +1,4 @@
 psutil==5.8.0
-pyroclient==0.1.1
+pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
 python-dotenv==0.17.0
 requests==2.25.1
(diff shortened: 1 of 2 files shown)


MateoLostanlen:
HI!, Thanks for this PR. Maybe you can also update [this requirement](https://github.com/pyronear/pyro-engine/blob/201234cc7b9896835a89bac771aee1e39cf3e8e5/server/src/requirements.txt#L9) 
The rooter library must be in the main.rpi.requirements.txt instead

frgfm:
@MateoLostanlen I only modified the files I felt like I understood where/when it's being used!
I'd need some guidance, because deps are not the same for central Pi vs camera controllers

MateoLostanlen:
ok for doing this in another PR, all good then !

MateoLostanlen (approved):

PR #62, opened 2021-09-15
reduce image size to save bandwidth

MateoLostanlen:
The connection is quite bad at the top of the towers, I propose here to send an HD image (960, 720) instead of the maximum resolution (3280, 2464)


fe51 (commented):
Do you think the frame size should be set up according to sites ? As long as sites does not all haves the same bandwith ? But that could be a later update. 

(However, not having same size for all frames could imply rework on front side )

frgfm (commented):
I think we should investigate impact on performances if this data can be used later on for training or evaluation. I agree that it helps with the bandwidth but at the same time, considering we don't use it much we would be decreasing the size of available images for firemen and other services

MateoLostanlen:
This change has already been made on the devices because we had bandwidth problems at the beginning of the summer. Then we can upload the backup images during the night now that we have the ability to do delayed uploads

pyroengine/engine/engine.py
```diff
@@ -90,6 +90,8 @@ def predict(self, frame, pi_zero_id=None):
                 self.event_appening[pi_zero_id] = True
 
             if self.use_api and self.event_appening[pi_zero_id]:
+                # Reduce image size to save bandwidth
+                frame = frame.resize(((960, 720)))
```
frgfm, line 94: Perhaps we should set the compressed size as a constructor argument (and its class attribute counterpart) So that if we change this in the future, we only have to pass another size when instantiating the engine

pyroengine/engine/engine.py
```diff
@@ -109,6 +111,8 @@ def predict(self, frame, pi_zero_id=None):
             if self.frames_counter[pi_zero_id] == self.save_evry_n_frame:
                 # Reset frame counter
                 self.frames_counter[pi_zero_id] = 0
+                # Reduce image size to save bandwidth
+                frame = frame.resize(((960, 720)))
```
frgfm, line 115: same here!

frgfm (commented):
Looks good to me! Only one suggestion: making the size a class attribute to easily change this

MateoLostanlen:
@frgfm you are right about the class argument, it will be way more flexible. I have made the change

pyroengine/engine/engine.py
```diff
@@ -102,6 +105,10 @@ def predict(self, frame: Image.Image, pi_zero_id: Optional[int] = None) -> float
             self.heartbeat(pi_zero_id)
             logging.info(f"Wildfire detection score ({prob:.2%}), on device {pi_zero_id}")
 
+        # Reduce image size to save bandwidth
+        if self.frame_size:
```
frgfm, line 109:
```python
if isinstance(self.frame_size, tuple):
```
would be more robust

pyroengine/engine/engine.py
```diff
@@ -142,6 +149,9 @@ def predict(self, frame: Image.Image, pi_zero_id: Optional[int] = None) -> float
             if self.frames_counter[pi_zero_id] == self.frame_saving_period:
                 # Reset frame counter
                 self.frames_counter[pi_zero_id] = 0
+                # Reduce image size to save bandwidth
+                if self.frame_size:
```
frgfm, line 153: same here

pyroengine/engine/engine.py
```diff
@@ -142,6 +149,9 @@ def predict(self, frame: Image.Image, pi_zero_id: Optional[int] = None) -> float
...
+                    frame = frame.resize(((960, 720)))
```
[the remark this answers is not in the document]
MateoLostanlen: hahaha good catch thanks :)

fe51 (approved):

frgfm (approved):
Looks good to me, thanks!

PR #63, opened 2021-09-15
Switch to script

MateoLostanlen:
This modification has been done on the devices but not on the repo, we need a script to reboot the rooter with a cronjob


fe51 (approved):
Hi Matéo! Thanks for the PR

frgfm (approved):
Thanks! I think you need to merge master & resolve conflicts, then we're good to go :)

PR #64, opened 2021-09-15
reorganize pi_utils

MateoLostanlen:
The "pi_utils" folder contains everything that is useful for the raspberry pi, it is better to separate what is utils to the pi zero from what is util to the main pi. In particular for the updates


pyroengine/pi_utils/.env
```diff
@@ -0,0 +1,3 @@
+ROOTER_LOGIN = "ROOTER_LOGIN"
+ROOTER_PASSWORD = "ROOTER_PASSWORD"
+ROOTER_IP = "ROOTER_IP"
```
frgfm, line 3: missing EOF

frgfm: And I would argue that this should be put in gitignore instead and we should add instructions to specify where this information should be added :thinking:

pyroengine/pi_utils/pi_zero/.env
```diff
@@ -0,0 +1,2 @@
+WEBSERVER_IP = "pi_central_hostname.local"
+WEBSERVER_PORT = "8002"
```
frgfm, line 2: same here

frgfm (commented):
Thanks! I added some comments, there are some conflicts to be resolved also here

MateoLostanlen:
I will delete these .env files and create a PR to add in the README steps to launch the system

frgfm:
Nice! There is one unittest failing though
And I suggest adding `.env` to the gitignore to avoid having people commit their file later on :)

PR #65, opened 2021-09-18
Improved router reboot script

MateoLostanlen:
This script to solve connection problems does not restart the rooter. We saw that the majority of the connection problems come from the rooter it is therefore essential to restart it in case of connection loss


frgfm (approved):
Thanks!

fe51 (approved):

PR #66, opened 2022-02-03
Fix docker

MateoLostanlen:
This PR aims to solve several docker problems in the repo. 
Pyro-engine Docker

1. The Dockerfile for pyro-engine was using the wrong image of pyro-vison
2. We have a problem with the version of pyro-vision (I will open an issue) so we have to delete the version for the moment

Web Server Docker

1. We delete the useless lines in the Dockerfile
2. We delete the useless huawei dependency and we don't block the version which can lead to problems later


server/src/requirements.txt
```diff
@@ -1,9 +1,8 @@
...
+uvicorn>=0.11.1
+python-multipart>=0.0.5
+pyrovision
+python-dotenv >= 0.17.0
 pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
-python-jose==3.2.0
-passlib[bcrypt]==1.7.4
-huawei-lte-api == 1.4.18
+python-jose>=3.2.0
+passlib[bcrypt]>=1.7.4
```
frgfm, line 8: That is a bit risky: I understand for provision that it was the only solution, but for the rest, since this is production environment, and we've encountered version issues in the past, I'd suggest sticking to full version pinning :sweat_smile:

server/src/Dockerfile
```diff
@@ -18,29 +18,7 @@ ENV PATH /usr/local/bin:$PATH
...
-RUN set -ex \
-	 \
-	&& savedAptMark="$(apt-mark showmanual)" \
-	&& apt-get update && apt-get install -y --no-install-recommends \
-		gcc \
-		python3-dev \
-		libffi-dev \
-		libssl-dev \
-	&& pip install --upgrade pip setuptools wheel \
-    && pip install -r /usr/src/app/requirements.txt \
```
frgfm, line 43: Is this because pyro-engine image already has all this?

frgfm (commented):
Thanks! Added a few questions/comments!

server/src/requirements.txt
MateoLostanlen: ok I get your point

server/src/Dockerfile
MateoLostanlen: Yes :)

server/src/requirements.txt
MateoLostanlen: Done !

requirements.txt
```diff
@@ -1,5 +1,5 @@
-pyrovision >= 0.1.2
-python-dotenv >= 0.15.0
-requests>=2.25.1
+pyrovision
```
frgfm, line 1: At least a lower bound would be nice don't you think?

server/src/requirements.txt
```diff
@@ -1,9 +1,8 @@
 fastapi==0.61.1
 uvicorn==0.11.1
 python-multipart==0.0.5
-pyrovision == 0.1.2
+pyrovision
```
frgfm, line 4: same comment here

frgfm (commented):
Just missing one version specifier for pyrovision I guess :)

requirements.txt
MateoLostanlen: We have an issue with our version number, we get `pyrovision--version-.-.-0.1.2a0-288795a` therefore the condition `pyrovision>xx` will always fail

server/src/requirements.txt
MateoLostanlen: same

requirements.txt
frgfm: I think this will fix the problem: https://github.com/pyronear/pyro-vision/pull/133

MateoLostanlen: Thanks ! But I was thinking, maybe it's better to install provision from the repo. We need all the latest updates

frgfm: Those are parallel topics (means of installation vs. version specifier), but I'd say:
- we need a version specifier for production repo :sweat_smile: 
- regarding the means of installation, installing from the repo means that if a bug slips in a commit, it will have it. if we put the normal one, we can always install from the repo before running that installation and have the choice. We'll need to make a release of pyrovision soon anyway 

What do you think?

MateoLostanlen: Yes, I see what you mean. I think we need to make a 0.1.3 release and define this version as a minimum to include recent bug fixes.

MateoLostanlen: can you review the [open_fire_v2 PR](https://github.com/pyronear/pyro-vision/pull/121) to include it in the release ?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 pyrovision@git+https://github.com/pyronear/pyro-vision.git#egg=pyrovision
-python-dotenv >= 0.15.0
-requests>=2.25.1
 pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
-psutil
+python-dotenv == 0.15.0
```
frgfm, line 3: can you confirm it can accept spaces around the double equal sign?

frgfm (approved):
Thanks! I'd only recommend to check that it's OK to have spaces around equal signs in the requirements.txt

requirements.txt
MateoLostanlen: it seems to work but I will remove it just in case

PR #69, opened 2022-02-13
Run by cron

MateoLostanlen:
This script allows to capture an image every 30s on the pi zero and send it to the central pi

For the moment it is launched in background using `python3 /home/pi/pi_zeros/runner.py &` but it is not very stable

I suggest to use a crontab instead which will launch a loop of two captures every minute:

`*/1 * * * * python3 /home/pi/pi_zeros/runner.py --max_iteration 2`

### DIFF
--- a/pyroengine/pi_utils/pi_zeros/runner.py
+++ b/pyroengine/pi_utils/pi_zeros/runner.py
@@ -53,13 +53,12 @@ class Runner:
 
     def run(self):
         it = 0
-        while self.is_running:
+        while True:
             it += 1
             files = self.capture_stream()
             self.send_stream(files)
-            if self.max_iteration is not None:
-                if it >= self.max_iteration:
-                    break
+            if self.max_iteration > 0 and it >= self.max_iteration:
+                break
 
             time.sleep(self.loop_interval)  # Wait between two captures
 
@@ -69,7 +68,7 @@ if __name__ == "__main__":
     parser.add_argument('--single', action='store_true', help='Single picture instead of eternal loop')
     parser.add_argument('--write', action='store_true', help='Use /write_image route instead of /inference')
     parser.add_argument('--loop_interval', default=30, action='store_true', help='Time between two photos')
-    parser.add_argument('--max_iteration', default=None, action='store_true', help='Time between two photos')
+    parser.add_argument('--max_iteration', default=0, action='store_true', help='Number of loop')
     args = parser.parse_args()
 
     webserver_local_url = f"http://{WEBSERVER_IP}:{WEBSERVER_PORT}" + \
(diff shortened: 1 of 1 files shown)


pyroengine/pi_utils/pi_zeros/runner.py
```diff
@@ -52,24 +52,29 @@ def send_stream(self, files):
...
-    def stop_runner(self):
-        self.is_running = False
+            time.sleep(self.loop_interval)  # Wait between two captures
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Take picture(s) and send to local web server')
     parser.add_argument('--single', action='store_true', help='Single picture instead of eternal loop')
     parser.add_argument('--write', action='store_true', help='Use /write_image route instead of /inference')
+    parser.add_argument('--loop_interval', default=30, action='store_true', help='Time between two photos')
```
fe51, line 70:
```suggestion
    parser.add_argument('--loop_interval', default=30, action='store_true', help='Time in seconds between two photos')
```

pyroengine/pi_utils/pi_zeros/runner.py
```diff
@@ -24,14 +24,14 @@ class Runner:
 
     logger = logging.getLogger(__name__)
     CAPTURE_DELAY = 3
-    LOOP_INTERVAL = 3
 
-    def __init__(self, webserver_url):
+    def __init__(self, webserver_url, loop_interval=30, max_iteration=None):
```
fe51, line 28:
```suggestion
    def __init__(self, webserver_url, loop_interval_s=30, max_iteration=None):
```

pyroengine/pi_utils/pi_zeros/runner.py
```diff
@@ -24,14 +24,14 @@ class Runner:
 
     logger = logging.getLogger(__name__)
     CAPTURE_DELAY = 3
```
fe51, line 26:
```suggestion
    CAPTURE_DELAY_S = 3
```

pyroengine/pi_utils/pi_zeros/runner.py
```diff
@@ -24,14 +24,14 @@ class Runner:
...
-    LOOP_INTERVAL = 3
```
fe51, line 27: LOOP_INTERVAL_S = 3

fe51 (changes requested):
Hi, thanks for the PR !
 
How did you spot the instability of the current process? In any case, the proposed approach via a crontab sounds good to me !

I have juste made some naming suggestion. If for variable naming suggestion, it is open to discussion (name might be too long now haha,  or a comment to make it clear it is about seconds ?), I think it's good to specify the time unit in the parser desc.

MateoLostanlen:
Hi @fe51, thanks for your comments. Yes you are right it's better to specify time unit. I've made the change

I don't really know why the background task it's unstable but I notice that after a while the process stop :/

fe51 (approved):
Cool, thanks fo the quick updates !

In fact, it occured  to me that I had also seen on forums that it was perhaps not stable..
 who knows why !
 I do not know why, but as long as you've experienced that it's not stable, and regular short job cron runs seem to be more robust, let’s go for it !

MateoLostanlen:
I'm closing this PR, let's use #74 instead

PR #72, opened 2022-03-13
chore: Installed pyrovision from source

MateoLostanlen:
Let's install pyro-vision from repo to keep it uptodate !


frgfm (approved):
Thanks!

PR #74, opened 2022-03-15
add monitor runner

MateoLostanlen:
The aim of this PR is to replace the PR #69, this solution allows thanks to a bash script to keep the runner alive.

This script must be added to the crontab to be launched at reboot time

@reboot bash /home/pi/pi_zeros/monitor_runner.sh


frgfm (approved):
Thanks!

PR #75, opened 2022-03-18
change default loop interval

MateoLostanlen:
The default  time between two frames should be 30s and not 3


frgfm (approved):
Thanks! Do you mean we were taking pictures every 3 seconds until now?

fe51 (approved):
Thanks for the update !

MateoLostanlen:
> Thanks! Do you mean we were taking pictures every 3 seconds until now?

No I've made the change by hand :) And I forget to right this PR

PR #77, opened 2022-07-14
Feature/switch to detection model

MateoLostanlen:
The goal of this PR is to integrate the use of the new object detection model using ultralytics repo. This allows to remove the dependency to pyro-vison for the moment

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,7 @@
-pyrovision@git+https://github.com/pyronear/pyro-vision.git#egg=pyrovision
 pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
 python-dotenv==0.15.0
 requests==2.25.1
 psutil==5.9.0
+torch
+torchvision
+opencv-python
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,5 +1,7 @@
-pyrovision@git+https://github.com/pyronear/pyro-vision.git#egg=pyrovision
 pyroclient@git+https://github.com/pyronear/pyro-api.git#egg=pyroclient&subdirectory=client
 python-dotenv==0.15.0
 requests==2.25.1
 psutil==5.9.0
+torch
+torchvision
+opencv-python
```
frgfm, line 7: please pin those versions :pray: 

Do we need openCV by the way?

MateoLostanlen: yes we do for know because we are using ultralytics to load the model but I will fix this in an upcoming PR ;)

MateoLostanlen:
Ok this PR this ready for review. As I said [here](https://github.com/pyronear/pyro-engine/pull/77#discussion_r921268341) there is a lot of dependencies due to ultralytics. I will make a PR soon to fix this.

@frgfm I will fix #78 when this PR will be merge !

frgfm (approved):
Thanks!

FYI, it's better not to mix code formatting with feature implementation :sweat_smile: Here it's a bit hard to notice the actual differences since there are some formatting changes (better in a separate PR)

PR #78, opened 2022-07-14
Feature/switch to reolink cam

MateoLostanlen:
The purpose of this PR is to adapt pyro-engine to our camera change. We are now going to use reolink ip cameras.

Previously the central raspberry pi was behaving like a server with an api, the ip cameras sent every 30s an image to this api for analysis.

Now the cameras are passive and it's the central pi that will fetch the images for analysis.


runner/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  pyro-engine:
+    build:
+      context: ..
+      dockerfile: ./runner/src/Dockerfile
+    command: python runner.py
+    volumes:
+      - ./data:/usr/src/app/data
+
+    restart: always
```
frgfm, line 12: Considering the simplicity of this, we might consider in the future to run it as a simple docker (not a docker compose)

runner/src/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM pyro-engine
```
frgfm, line 1: What is this expected to do? Shouldn't we pull from dockerhub?

frgfm (commented):
Thanks for the refactor !

Would you mind:
- moving server/data out of this empty folder
- moving runner/docker-compose.yml to root
- moving runner/src to src
- updating / fixing the unittests

That will give us a much cleaner project :)

runner/docker-compose.yml
MateoLostanlen: We can leave it as it is for know and discuss it soon :)

MateoLostanlen:
I will improve coverage when #83 will be merge

MateoLostanlen:
The loss of coverage here is on engine but you would have to simulate an api or create test creds to test it. I propose to do it in a next PR

MateoLostanlen:
Let's merge this

frgfm:
> I will improve coverage when #83 will be merge

oops sorry, I thought you were saying we should wait for #83 to review!

frgfm (approved):
Cheers!

PR #80, opened 2022-07-18
Drop useless files

MateoLostanlen:
The aim of this PR is to do a bit of cleaning:

1. we are not using pizeros so pizeros related files are removed
2. We are not usiing a wifi routeur so routeur related files / elements are removed too
3. Drop useless one test


MateoLostanlen:
I will improve coverage when #83 will be merge

pyroengine/pi_utils/check_internet_connection.sh
```diff
@@ -36,7 +36,5 @@ then
     exit 0;
 fi;
 
-python3 /home/pi/pyro-engine/pyroengine/pi_utils/reboot_router;
-sleep 60;
```
frgfm, line 40: no longer relevant?

frgfm (commented):
Thanks! I wasn't sure about the router reboot check though

pyroengine/pi_utils/check_internet_connection.sh
MateoLostanlen: no this was to reboot the wifi router that we are not using anymore

frgfm (approved):
Thanks!

PR #84, opened 2022-07-19
fix version

MateoLostanlen:
Ultralytics [require](https://github.com/ultralytics/yolov5/blob/92e47b85d952274480c8c5efa5900e686241a96b/requirements.txt#L14) tqdm >= 4.64.0


frgfm (approved):
Thanks!

PR #93, opened 2022-07-24
Bugfix/get image

MateoLostanlen:
I did not really understand why but the request with the url in http bugs sometimes which is not the case in https


MateoLostanlen:
I'm resetting a new pi-runner so it's not working right now

PR #95, opened 2022-07-24
Feature/capture

MateoLostanlen:
As discussed offline we need a function to capture an image from each camera mainly for debugging purposes. However it is dangerous to do what was done in the PR #92 we don't want to interfere with the runner. This PR proposes to remove the modification from the previous PR and create an independent script


frgfm:
Quick question: why did you add the label "invalid" ? :sweat_smile:

MateoLostanlen:
Hi @frgfm, the current capture method was "invalid". But maybe I'm not using the tag properly

frgfm:
capture method? what are we talking about? the camera?

the tag "invalid" is usually used on "invalid" issues for instance :sweat_smile:  (meaning that the raised concern / issue is irrelevant or non applicable)

PR #98, opened 2022-07-26
Feature/remove ultralytics

MateoLostanlen:
The goal of this PR is to get rid of the ultralytics repo and all its dependencies with a class that allows to perform inference directly from the onnx model


pyroengine/models/utils.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (C) 2020-2022, Pyronear.
```
frgfm, line 1:
```
# Copyright (C) 2022, Pyronear.
```

(the file / code wasn't created in 2020)

requirements.txt
```diff
@@ -5,12 +5,5 @@ psutil==5.9.0
 # From https://github.com/ultralytics/yolov5/blob/master/requirements.txt
 torch==1.11.0
 torchvision==0.12.0
-opencv-python==4.5.4.60
-pandas==1.1.4
-PyYAML==5.3.1
-Pillow==9.2.0
-tqdm==4.64.0
 onnx==1.12.0
```
frgfm, line 8: we don't need onnx, only onnxruntime

frgfm: however, we need numpy (yes it's included in torch / torchvision but it gets removed from this, we need to specify all direct dependencies)

pyroengine/engine/predictor.py
```diff
@@ -3,29 +3,24 @@
...
 
-import torch
+from pyroengine.models.yolov5 import Yolo_v5
 
 
 class PyronearPredictor:
     """This class use the last pyronear model and run our smoke detection model on it
     Examples:
-        >>> pyronearPredictor = PyronearPredictor()
+        >>> pyronearPredictor = PyronearPredictor(model_weights=/path/to/model.onnx)
```
frgfm, line 12:
```
>>> pyronearPredictor = PyronearPredictor(model_weights="/path/to/model.onnx")
```

pyroengine/models/yolov5.py
```diff
@@ -0,0 +1,47 @@
...
+            model_weights (str): path to onnx file
+            conf_thres (float, optional): confidence threshold. Defaults to 0.25.
+        """
+        self.session = ort.InferenceSession(
+            model_weights, providers=["CPUExecutionProvider"]
+        )
+
+        self.conf_thres = conf_thres
+
+    def forward(self, im):
```
frgfm, line 29: this isn't a Python module, so we should implement
```python
def __call__(self, im):
```

pyroengine/engine/predictor.py
```diff
@@ -3,29 +3,24 @@
...
-        self.model.conf = conf
+        self.model = Yolo_v5(model_weights=model_weights, conf_thres=conf)
 
     def predict(self, im):
         """Run prediction"""
-        pred = self.model(im)
+        pred = self.model.forward(im)
 
-        return pred.xyxy[0].numpy()
+        return pred
```
frgfm, line 26: According to the constructor and the call method implementation, this class does nothing that `Yolo_v5` doesn't

I'd suggest directly instantiating a `Yolo_v5`

pyroengine/models/yolov5.py
```diff
@@ -0,0 +1,47 @@
...
+        """Run prediction
+
+        Args:
+            im (Pillow): Image to analyse
+
+        Returns:
+            np.array: prediction with [x1, y1, x2, y2, conf, class] where xy1=top-left, xy2=bottom-right
+        """
+        im = im.resize((640, 640))
+        x = np.expand_dims(np.array(im).astype("float"), axis=0)
```
frgfm, line 39: I don't suggest inplace modification here
```python
x = np.asarray(im.resize(self.size), dtype=np.float32)[None, ...]
```

pyroengine/models/yolov5.py
```diff
@@ -0,0 +1,47 @@
...
+        x = np.ascontiguousarray(x.transpose((0, 3, 1, 2)))  # BHWC to BCHW
+        x = torch.from_numpy(x).float() / 255  # uint8 to fp16/32
+
+        y = self.session.run(["output"], {"images": x.numpy()})[0]
+
+        output = non_max_suppression(torch.tensor(y), self.conf_thres)[0].numpy()
```
frgfm, line 45: do we need to convert it back to a pytorch tensor?

Can't we do it with numpy? I'd suggest implementing the utils in pure numpy to get rid of PyTorch & torchvision. (onnx works with numpy either way, not Pytorch tensors)

frgfm (commented):
Thanks a lot!

I think we could do a few modifications to get something really neat :)

MateoLostanlen:
This is not relevant anymore after #101

PR #101, opened 2022-08-03
refactor: Cleaned pyroengine and refactored project

frgfm:
This PR introduces the following changes:
- simplified `pyroengine`
    - deprecates `pi_utils` (not used)
    - moved `pyroengine.engine.engine` to `pyroengine.core`
    - added `pyroengine.vision` with `Classifier` which holds the ONNX model
- updated `tests`
    - moved "test" to "tests"
    - switched from unittests to pytest to harmonize our test suite across repos
    - added exhaustive unit tests for pyroengine (only part not being tested : API interactions, because it requires to run it locally)
- updated `docs`
    - updated documentation theme from rtd_theme to furo (cf. modifications in pyrovision)
- deleted `server` folder
- renamed `runner` as `src`
- cleaned `src` by implementing `ReolinkCamera` and `SystemController`
- refactored all requirements.txt and config files into `pyproject.toml`
- added a `Makefile`
- updated CI jobs accordingly
- updated docker file accordingly

Basically, for the wrapper (orginially `runner`, now `src`), the only difference is:

- previously, we needed the following keys: "model_weights"
- now, we need: "cfg_path" (describing the preprocessing), "model_path" (path to ONNX model), "hub_repo" (fallback to download the model from HF hub)

For reviewers, I highly suggest to check the modifications by commit, it might be easier :sweat_smile:


pyproject.toml
```diff
@@ -0,0 +1,128 @@
...
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Mathematics",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+]
+dynamic = ["version"]
+dependencies = [
+    "Pillow>=8.4.0",
```
MateoLostanlen, line 33: it's not better to use exact versions for deployment ?

pyroengine/core.py
```diff
@@ -0,0 +1,292 @@
...
+            self._dump_cache()
+            self.last_cache_dump = ts
+
+        # save frame
+        if len(self.api_client) > 0 and isinstance(self.frame_saving_period, int) and isinstance(cam_id, str):
+            self.frames_counter[cam_id] += 1
+            if self.frames_counter[cam_id] == self.frame_saving_period:
+                # Send frame to the api
+                stream = io.BytesIO()
+                frame.save(stream, format="JPEG")
```
MateoLostanlen, line 228: maybe it's not the time to do it but jpeg default compression is 75% for jpeg, it's a bit low

.github/workflows/docker.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches: main
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
```
MateoLostanlen, line 11: The goal here is to build an image to test on your pc ? You are crushing the image used by the towers that must be built for arm. We have to setup a cross-build but we can't overwrite the existing image

pyroengine/__init__.py
```diff
@@ -1,6 +1,2 @@
-# Copyright (C) 2020-2022, Pyronear.
```
MateoLostanlen, line 1: Why do you remove the header here ?

MateoLostanlen (changes requested):
Hi @fg, thanks so much for this PR! Globally LGTM. Several sticking points though:

1. This PR must be merge on develop and not main directly
2. CIs must run on main & develop

To discuss

We never used the felix code to monitor the pi's but I think it would be nice to do so, I don't know if it's a good idea to delete it

NB: It would be better to use a fork

pyproject.toml
```diff
@@ -0,0 +1,128 @@
...
+keywords = ["pytorch", "deep learning", "vision", "cnn", "wildfire"]
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: Apache Software License",
+    "Natural Language :: English",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.6",
```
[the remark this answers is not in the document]
frgfm: Well the only thing constraining the version of Python in Python3 is the use of f-string
Subdependencies might have subtilities, but our codebase is fully compatible with 3.6 & higher :)

pyproject.toml
frgfm: Yes it is, but I figured the package should have loose specifiers, and `src/requirements.txt` should have fixed ones :)

pyroengine/core.py
frgfm: Mmmh the option to downsize the image is already in place, we can check this in another PR (since we'll need some benchmarking)

pyroengine/__init__.py
frgfm: files with no added value code (`__init__.py` and `version.py`) have no reason to include the header 
It's just adding commented sections in files that have barely anything

.github/workflows/docker.yml
frgfm: ? I'm not sure what you mean: I haven't removed anything here, only added a test.

To me, we have to make sure that the image can at least be built on ubuntu. If it doesn't, we can expect it to fail on ARM. Now, I wanted to add also a CI to build the image on ARM but I have no clue how long it takes on RPI. I suggest to add this in another PR along with the orchestration of the CI to include API interactions :+1:

pyproject.toml
frgfm: Actually, let me set poetry in place for the `src`, that will help us

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          python-version: 3.8
+          architecture: x64
+      - name: Install package
+        run: |
+          python -m pip install --upgrade pip
+          pip install pyroengine
+          python -c "import pyroengine; print(pyroengine.__version__)"
+
+  dockerhub-publish:
```
MateoLostanlen, line 58: My bad the push is here

.github/workflows/docker.yml
MateoLostanlen: You erase the [current image](https://hub.docker.com/repository/registry-1.docker.io/pyronear/pyro-engine/tags?page=1&ordering=last_updated) on hub but pushing a linux/arm64 image, but the docker pushis not here sorry !

MateoLostanlen: https://github.com/pyronear/pyro-engine/blob/ea017e1cff156f6d1137cd7684180288b7aa8a21/.github/workflows/release.yml#L58

MateoLostanlen: But yes we can fix it manually and managed this in a comming PR

pyproject.toml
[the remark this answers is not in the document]
MateoLostanlen: ok ok

pyroengine/core.py
MateoLostanlen: Yes I just wanted to raise the "issue", we can manage this later with a proper bench

pyproject.toml
frgfm: will be resolved in #102

MateoLostanlen:
> Hi @fg, thanks so much for this PR! Globally LGTM. Several sticking points though:
> 
>     1. This PR must be merge on develop and not main directly
> 
>     2. CIs must run on main & develop
> 
> 
> To discuss
> 
> We never used the felix code to monitor the pi's but I think it would be nice to do so, I don't know if it's a good idea to delete it
> 
> NB: It would be better to use a fork

Thanks for the update, what about these comments ?

frgfm:
> Thanks for the update, what about these comments ?

Oops I didnt see the last part:
- I removed it because it comes with dependencies (I would have kept it otherwise)
- thought it would be better to start with a clean environment, then add it back later and see how we integrate it

MateoLostanlen (approved):
LGTM

PR #106, opened 2022-08-05
ci: Added multi-arch docker build upon release

### LINKED ISSUE #103: [docker] Docker build management
Hey there :wave: 

As discussed offline, we need to handle docker containers efficiently for this repo as it is critical for deployment.
Here is my suggestion:
- it is mandatory that the DockerHub repo https://hub.docker.com/repository/docker/pyronear/pyro-engine has the latest ARM images.
- CI docker push frequency: either we push at each merged PR, or at each release. I suggest doing those on releases, we can increase the release pace to use patch index (i.e. `0.1.1`, `0.1.2`, etc.)
- tag management : latest version should be pushed to the tag `latest`. Additionally, it should also be pushed to a descriptive tag: for instance, as of right now, the image is based on python:3.8.1-slim, so the tag should be `python3.8.1-slim`.
- device update: once pushed to dockerhub, a digest hash is created. This should be the reference for the software hash of the device.
- optionally, if we can manage to push to DockerHub the linux build as well (not overriding the ARM one), that would be nice :+1: 

Any suggestion is welcome!

frgfm:
This PR adds a CI job to build and push docker images for both linux/amd64 and linux/arm64, following this tutorial: https://dev.to/cloudx/multi-arch-docker-images-the-easy-way-with-github-actions-4k54

Closes #103 

Any feedback is welcome!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,15 +56,26 @@ jobs:
           python -c "import pyroengine; print(pyroengine.__version__)"
 
   dockerhub-publish:
+    if: "!github.event.release.prerelease"
     name: Push Docker image to Docker Hub
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to Docker Hub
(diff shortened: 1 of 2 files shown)


MateoLostanlen (approved):
Indeed it is much easier with buildx ! did you test it ?

frgfm:
@MateoLostanlen not yet, it will be triggered on the next release!
If it fails, I'll trigger it manually and fix the job :ok_hand:

MateoLostanlen:
Ok thanks !

PR #107, opened 2022-08-05
ci: Added conda build & publish job for releases

frgfm:
This PR adds a CI job to build and publish to conda on each release.

Any feedback is welcome!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,55 @@ jobs:
           pip install pyroengine
           python -c "import pyroengine; print(pyroengine.__version__)"
 
+  conda-publish:
+    if: "!github.event.release.prerelease"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Miniconda setup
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -55,6 +55,55 @@ jobs:
...
+          ls -l conda-dist/noarch/*tar.bz2
+          anaconda upload conda-dist/noarch/*tar.bz2
+
+  conda-check:
+    if: "!github.event.release.prerelease"
+    needs: conda-publish
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
MateoLostanlen, line 92: maybe we can add self-hosted, what do you think?

frgfm: Well I don't think we'll use conda on Raspberry pi?
conda is more for multi env management (so for developers mostly, not really for production)

MateoLostanlen: Yes you are right ! all good then

MateoLostanlen (approved):

PR #109, opened 2022-08-06
ci: Added docker build CI job on self-hosted

frgfm:
This PR adds self-hosted as runner for the docker build job of the CI.

Any feedback is welcome!


frgfm:
Nevermind, it looks like it would need more configuration! It's alright if this is only performed on releases

PR #117, opened 2022-09-06
fix: Updated config files location in capture.py

fe51:
In capture.py, paths and name of configuration files are from the old architecture of the repo (before the last release)

This short PR introduces the following modifications: 
- in capture.py : updates path to .env 
- in capture.py: updates name and path to camera credentials

Happy to discuss it !

### DIFF
--- a/src/capture.py
+++ b/src/capture.py
@@ -28,7 +28,7 @@ def capture(ip, CAM_USER, CAM_PWD):
 
 
 def main():
-    load_dotenv("/home/pi/pyro-engine/runner/data/.env")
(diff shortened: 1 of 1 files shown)


frgfm:
Just FYI: I did update all scripts but the capture one so far :sweat_smile: 
So it actually needs some rework

fe51:
Hehe, I don't doubt it given the quality of the rest :D

This quick fix makes it functional, but I agree a rework has to be done !

frgfm (approved):
If this modification gets it to work, let's move forward with it :)

PR #118, opened 2022-09-07
feat: Added log management

fe51:
The size of the logs is not yet a problem, but on the one hand we have quite a lot of logs (especially in case of trouble), and on the other hand the current approach is to simply make the log file grow infinitely ♾️  !

Hence, in order to prevent possible future issues, we should manage log files rotation and remove oldest. 

Fortunately, this can be done in a few lines with [docker](https://docs.docker.com/config/containers/logging/json-file/). 

This PR introduces the following modifications : 
- Adds logging to docker compose (json-file logging driver, size max and 5 file max)

As long as there space available on devices, I've picked max-size as 100mo and 5 file maximum but open to discuss it :) 

Happy to discuss it 😄 

PS : log files are accessible at `/var/lib/docker/containers/the_relevant_container/`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,3 +13,8 @@ services:
     volumes:
       - ./data:/usr/src/app/data
     restart: always
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "100m"
+        max-file: "5"


fe51:
- "_about the logs of the docker, I would question the usefulness of dumping the logs in a static file_" --> In fact, according to my understanding of the [doc](https://docs.docker.com/config/containers/logging/json-file/) 
> By default, Docker captures the standard output (and standard error) of all your containers, and writes them in files using the JSON format." 
Hence, logs are already stored as static files. The idea here is to able to manage size file and avoid an exessive size, by defining a maximum size (having multiple files allows rotation (oldest is removed when max file is reached))

- "regarding the logs, it's better to host them on papertrail or logtail, similarly to the API & the platform" --> We have to be careful in our data usage of our internet connection. But eventually, I agree, for some logs (to define, because some would be redondant with info send to the api, and not useful) it would be nice to send and host them. 
Do you see it as a similar subject of hosting some metrics (temperature, cpu usage etc) of devices  ? (non only the ones from the docker containers)

MateoLostanlen (approved):
if the logs are already stored in static and this method allows to limit the size of the file it is rather a good thing

fe51:
@frgfm are you ok to merge this ?

frgfm (approved):
Sounds good!

frgfm:
@fe51 don't forget to delete source branch once the PR is merged ;)

PR #120, opened 2022-09-13
refactor: Updated src/capture.py

frgfm:
- refactors class definitions from src/run.py to src/controller.py
- updated src/capture.py to use those classes
- added JPEG 50 quality before saving/uploading pictures

Any feedback is welcome!


fe51:
First of all thanks for the PR !

Then some naive questions :)
why system controler.py is in src and not in pyroengine ?
why the arg is name "cache "to specify where to save snapshots ?

MateoLostanlen (changes requested):
Thank you for this refacto, it is much better like that. However, it doesn't work as it is, there is an import problem. I haven't looked into it in detail. Have you tried on the pi / in the docker?
I think it would be better to have the controller in the pyroengine package. What do you think ?

Also, can you update the docker accordingly? Especially with:

`COPY ./src/capture.py /usr/src/app/capture.py`

frgfm:
The last commits solve the import issue :+1:

MateoLostanlen (approved):
Sorry for this late review, all good for me ! Thanks

fe51 (approved):

PR #121, opened 2022-09-24
switch axis

MateoLostanlen:
Pillow images are define as W, H, C


frgfm (changes requested):
I tested this but there is a mistake: the channel order is not the problem, the resizing is :sweat_smile: I'm opening a PR to fix this!

MateoLostanlen:
Oh True, I made the confusion with the resize my bad !

MateoLostanlen:
done in #124

PR #124, opened 2022-09-25
fix: Fixes PIL resizing for non squared shapes

frgfm:
- fixes the resizing for non-squared images
- updates unittests expected shapes
- updates unittests model to save bandwidth

Any feedback is welcome!

### DIFF
--- a/pyroengine/vision.py
+++ b/pyroengine/vision.py
@@ -47,7 +47,7 @@ class Classifier:
         """
 
         # Resizing
-        img = pil_img.resize(self.cfg["input_shape"][-2:], Image.BILINEAR)
+        img = pil_img.resize(self.cfg["input_shape"][-2:][::-1], Image.BILINEAR)
         # (H, W, C) --> (C, H, W)
         img = np.asarray(img).transpose((2, 0, 1)).astype(np.float32) / 255
         # Normalization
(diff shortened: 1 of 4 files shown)


pyroengine/vision.py
```diff
@@ -47,7 +47,7 @@ def preprocess_image(self, pil_img: Image.Image) -> np.ndarray:
         """
 
         # Resizing
-        img = pil_img.resize(self.cfg["input_shape"][-2:], Image.BILINEAR)
+        img = pil_img.resize(self.cfg["input_shape"][-2:][::-1], Image.BILINEAR)
```
MateoLostanlen, line 50: Please use Image.Resampling.BILINEAR instead, Image.BILINEAR is [deprecated](https://pillow.readthedocs.io/en/stable/deprecations.html)

MateoLostanlen (changes requested):
Thanks for this PR, just on small change

pyroengine/vision.py
frgfm: This change would require also to update the version specifiers for Pillow, so I suggest doing it in another PR :+1: 

(either putting <9.1.0 in the specifier, or going for your change + >=X.Y.Z, where X.Y.Z is the first release where Resampling become supported)

MateoLostanlen: OK let's stay like this then

MateoLostanlen (approved):

PR #125, opened 2022-09-25
feat: Adds an option to specify the hash of the HF hub commit

### LINKED ISSUE #123: Model version management
I haven't used hugguing face much yet but I guess there is a way to track model evolutions. It should be used in Pyro-engine, the fact that a model change breaks the pipeline without our intervention is a problem.

frgfm:
This PR adds the option to use a specific commit of HF hub repo for file download. It also adds a case in the unittest.

Closes #123 

Any feedback is welcome!


MateoLostanlen (approved):
Thanks for this !

PR #126, opened 2022-10-02
split build

MateoLostanlen:
We modify much more often the package than the requirements, I think it is better to split the two installations in order to save time and avoid using too much of our internet credit


frgfm (commented):
Thanks!
Two open questions:
1. What is the difference in terms of Docker image size?
2. since you still have an installation step after the `COPY`, I thin that won't reduce the build duration too much. Have you noticed any big difference? (because if not, it's always better to minimize the number of `RUN` commands for image squashing)

MateoLostanlen:
Jump, no by emptying the cache the image size does not change 364Mb I tested.

Requirements installation is by far the longest part, especially if the connection is bad

PR #139, opened 2022-11-23
Feature/model managment

MateoLostanlen:
The purpose of this RP is to solve two of our current problems:

1. The too high number of false positives, for that I propose a kind of model ensemble by taking the average score of n models (here 2). I have done several performance tests on the subject, this technique allows to maintain almost unchanged the recall while improving significantly the precision. I'll share the results of my tests soon, I have to write this properly. The best combo among all the models available on the hub seems to be "pyronear/rexnet1_3x", "pyronear/rexnet1_5x" so I suggest to use it.

2. We have to limit our data consumption, so it is not optimal to download the models at each docker build. So I propose here to copy them in cache in the docker volume.

I could have done 2 PR but the two modifications are quite correlated so I left everything in the same one.

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -73,7 +73,7 @@ class Engine:
         """Init engine"""
         # Engine Setup
 
-        self.model = Classifier(model_list, **kwargs)
+        self.model = Classifier(model_list, cache_folder)
         self.conf_thresh = conf_thresh
 
         # API Setup
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -30,7 +30,7 @@ class Engine:
     """This implements an object to manage predictions and API interactions for wildfire alerts.
 
     Args:
-        hub_repo: repository on HF Hub to load the ONNX model from
+        models: list of models to use
         conf_thresh: confidence threshold to send an alert
         api_url: url of the pyronear API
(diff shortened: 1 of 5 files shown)


fe51 (approved):
Hello Mateo, thanks for the PR ! 

Its looks okay to me and I am going to quickly deploy it on St peray test site ! 

Maybe you could try to fix the tests !

frgfm:
My bad, I spoke with Mateo offline but better to document things here:
- while model ensembling is a good idea, even if we have RAM left, I think we shouldn't deploy this unless it solves 80-90% of our problems :/
- perhaps some things that could be done: knowledge distillation, student/teacher model, etc.

fe51:
> we shouldn't deploy this unless it solves 80-90% of our problems :/
--> Why not having this option ? 
--> So far, Mateo, do you have an idea of the improvement thanks to model ensembling ? 
 
Although for various reasons we would be better off taking another approach as you suggest, this solution improves our current alert system, so it is worth keeping as an option if we don't have something better by the time we get to maintenance in the Ardèche. (and that would help to stop spam from the st peray test tower)

MateoLostanlen:
I think the patch is worth it at least temporarily. I'll make available the results of my benchmark in the coming week to continue the debate

MateoLostanlen:
new model coming I'm closing this

PR #141, opened 2023-01-29
clean backup by size

MateoLostanlen:
Previously the local backup was clean after 30 days which could saturate the disk in case of a large number of false positives. So I propose to clean it according to its size

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -3,6 +3,7 @@
 # This program is licensed under the Apache License 2.0.
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0> for full license details.
 
+import glob
 import io
 import json
 import logging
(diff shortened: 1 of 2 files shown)


pyroengine/engine.py
```diff
@@ -26,6 +27,13 @@
 logging.basicConfig(format="%(asctime)s | %(levelname)s: %(message)s", level=logging.INFO, force=True)
 
 
+def get_folder_size(folder):
+    return (
+        sum(os.path.getsize(f) for f in glob.glob(str(folder) + "/**/*", recursive=True) if os.path.isfile(f))
+        // 1000000
+    )
+
+
```
frgfm, line 36: Let's do it more precisely : `// 1024 ** 2`
And only a suggestion but perhaps it's better to avoid creating a function for this (using the command directly in the method)

frgfm (commented):
Thanks! Looks good, I only added a small suggestion. And have you tested it?

pyroengine/engine.py
MateoLostanlen: yes good idea !

MateoLostanlen:
> Thanks! Looks good, I only added a small suggestion. And have you tested it?

yes it works :)

frgfm (approved):
Thanks! We'll need to fix the unittests at some point

PR #142, opened 2023-01-29
Day only

MateoLostanlen:
For now our models are not very efficient at night, so I propose not to raise any alert at this time.
In this PR I added a script that runs via crontab each night to get the sunrise and sunset times and a function that checks that we are in the daytime hours before processing an image

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -28,7 +28,7 @@ logging.basicConfig(format="%(asctime)s | %(levelname)s: %(message)s", level=log
 
 
 def is_day_time(cache):
-    """Read sunset and sunrise hour in sunset_sunrise.txt and check if we are currently on daytime. We don't want to 
+    """Read sunset and sunrise hour in sunset_sunrise.txt and check if we are currently on daytime. We don't want to
     trigger night alerts for now. We take 1 hour margin
 
     Args:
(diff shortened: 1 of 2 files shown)


pyroengine/engine.py
```diff
@@ -27,6 +27,33 @@
...
+    sunrise = lines[0][:-1].split(":")
+    sunset = lines[1][:-1].split(":")
+    th = datetime.now().hour
+    tm = datetime.now().minute
+    if (
+        th < int(sunrise[0]) - 1
+        or (th == int(sunrise[0]) - 1 and tm < sunrise[1])
+        or th > int(sunset[0]) + 1
+        or (th == int(sunset[0]) + 1 and tm > int(sunset[1]))
+    ):
```
frgfm, line 51: I suggest parsing the string with `datetime.strptime(date_string, format)` to get datetime comparison directly :)

pyroengine/engine.py
```diff
@@ -27,6 +27,33 @@
...
+        return False
+    else:
+        return True
```
frgfm, line 54: and here you can return `not` your condition to avoid the if

frgfm (commented):
Thanks! I added a few comments :ok_hand:

pyroengine/engine.py
MateoLostanlen: done :)

pyroengine/engine.py
```diff
@@ -27,31 +27,23 @@
...
-        or th > int(sunset[0]) + 1
-        or (th == int(sunset[0]) + 1 and tm > int(sunset[1]))
-    ):
-        return False
-    else:
-        return True
+    sunrise = datetime.strptime(lines[0][:-1], "%H:%M")
+    sunset = datetime.strptime(lines[1][:-1], "%H:%M")
+    now = datetime.strptime(datetime.now().isoformat().split("T")[1][:5], "%H:%M")
+    return (now - sunrise).total_seconds() > -delta and (sunset - now).total_seconds() > -delta
```
frgfm, line 46: I think there's a mistake :thinking: 
```suggestion
    return (now - sunrise).total_seconds() > delta and (sunset - now).total_seconds() > delta
```

frgfm (commented):
Not 100% sure, but I think there is a small mistake!

pyroengine/engine.py
MateoLostanlen: no if it's 8:15 and the sun rise at 8:20, (now - sunrise).total_seconds() = - 5*60 = -300, so we have -300 < -delta (-3600) therefore the system start

MateoLostanlen: I can change now and sunrise order if you prefer to avoid the -delta

MateoLostanlen: I want to start trigger alerts 1h before the sunrise because there is already day light usually, maybe the confusion come from that point

frgfm: Oh I thought you wanted 1h after sunrise :+1:

frgfm (approved):
Sounds good!

pyroengine/engine.py
MateoLostanlen: :) I thinks it's better this way, we can adapt this delta later if needed

PR #143, opened 2023-01-29
Yolov5

MateoLostanlen:
The purpose of this PR is to replace our detection model to use the last yolov5 that I trained.

I have a clear improvement in performance especially in terms of false positives. On the engie dataset (not present in the training ds) for example I go from 13% error for the rexnet1_5x or 14% for the rexnet1_x to less than 2.5%. On the DS ardeche (present partially in the training ds) I am <0.5%.

The predicitons are also quite good, I'll put some pictures [here](https://drive.google.com/drive/folders/1QxfNjzvlFm0wOEuBDSakgQ0jVQ5H8O_4?usp=share_link). There are the errors on the ardeche and engie ds and the predictions on my validation ds.

I have voluntarily chosen not to put the NMS code here to avoid unnecessary computation, but I have prepared a code if we want to exploit the location information soon


MateoLostanlen:
I'm a bit annoyed to have to reintroduce the dependency to opencv for my augmentation but the letterbox resize done with pillow doesn't produce the same thing as opencv (as shown [here](https://github.com/python-pillow/Pillow/issues/2718)) which has a significant impact on the prediction (I don't know exactly why they talk about it on the link above I have to dig into it). I propose to merge like this for the moment I will open an issue to deal with the problem

pyroengine/engine.py
```diff
@@ -242,7 +242,7 @@ def predict(self, frame: Image.Image, cam_id: Optional[str] = None) -> float:
             to_be_staged = self._update_states(pred, cam_key)
             if to_be_staged and len(self.api_client) > 0 and isinstance(cam_id, str):
                 # Save the alert in cache to avoid connection issues
-                self._stage_alert(frame, cam_id)
+                self._stage_alert(frame_resize, cam_id)
```
frgfm, line 245: If `self.frame_size` is not set, I think this is gonna crash :sweat_smile:

src/requirements.txt
```diff
@@ -24,3 +24,4 @@ sympy==1.10.1; python_version >= "3.7" and python_version < "4"
 tqdm==4.64.0; python_version >= "3.6" and python_version < "4" and python_full_version >= "3.7.0"
 typing-extensions==4.3.0; python_version >= "3.7" and python_version < "4" and python_full_version >= "3.7.0"
 urllib3==1.26.11; python_full_version >= "3.7.0" and python_version < "4" and python_version >= "3.7"
+opencv-python==4.5.5.64; python_full_version >= "3.7.0" and python_version < "4" and python_version >= "3.7"
```
frgfm, line 26: Have you checked that this version of OpenCV is running smoothly on ARM CPU?

frgfm (commented):
Sorry about the late review!

pyroengine/engine.py
MateoLostanlen: good catch I will fix it !

src/requirements.txt
MateoLostanlen: What do you mean by running smoothly ? Where I can check that ?

MateoLostanlen:
thanks your your review @frgfm, sorry I by-passed the approval I setting up the devices for tomorrow

src/requirements.txt
frgfm: OpenCV has some bindings to low level languages, which needs to be compiled (similarly to PyTorch). And since we don't want to build OpenCV for ARM (just like what we did for PyTorch in the early days), we need to make sure that this requirement is installing smoothly on ARM/Raspberry Pi (no error on install & basic command) compared to common CPU architectures

PR #144, opened 2023-01-30
Make all params availables from run

MateoLostanlen:
The purpose of this pr is to make all engine parameters accessible from run so it is easier to test different configurations.

Also I changed some default values.

frame_size 4k > 720,1280 the 4k image is too heavy for some towers

jpeg_quality 50 > 80, the compression was too strong with some artifacts especially in low light

cache-size 100 > 20 to try not to saturate the ram

alert_relaxation 3 > 2 because the model is better now I propose to decrease the alert_relaxation


fe51 (approved):
Good idea to have all params available as args

Issue #146, opened 2023-03-03
Workaround for the mysterious Docker failure


fe51:
We experience on our devices that the docker stops working, i.e. the docker is still running, is active, but nothing happens (no inference, no api interactions etc) 

Until now, since this bug has no root cause and appears after a random runtime, we have no robust solution. 

On the other hand, a simple docker stop, docker up (@MateoLostanlen if could explicit the exact commands   ), bring us back to the expected docker behavior. 

Therefore, in order to keep the main functionalities of our product, until we have a long-lasting solution, I suggest to implement a workaround as for example, to regularly execute the actions that lead us to review the correct functioning. 

Happy to discuss it !

MateoLostanlen:
solved by #153

PR #148, opened 2023-04-07
add relaunch script

MateoLostanlen:
As you know we have stability problems with docker which crashes regularly. I am still trying to solve this problem in the meantime I propose to restart the docker every day at 1am using a cronjob and this bash script


fe51 (approved):
As it is a quick win to address this issue #146 and it will stabilize the functioning of our systems, it's ok for me !

MateoLostanlen:
This is not needed anymore thanks to #153

PR #149, opened 2023-04-07
update header

MateoLostanlen:
update headers, the ci do not pass anymore

it's not practical if you have to do it by hand every year, we should automatize it


fe51 (approved):

fe51:
Indeed, might need to be automated if possible (not to long to update though)

PR #155, opened 2023-04-21
add utc

MateoLostanlen:
I forgeted to take utc into account in #142, however our raspberry pi are set to UTC 0. Therefore we need to convert sunrise / sunset hour to utc 0 too.


fe51:
Nice catch !

fe51 (approved):

PR #163, opened 2023-07-24
stop saving random

MateoLostanlen:
Saving random images became useless, it was a good idea at the beginning to grow our dataset but we have a lot of data now so let’s focus on important ones. Moreover we are moving slowly to temporal prediction so having one image alone will be useless

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -75,7 +75,6 @@ class Engine:
             the number of consecutive negative detections before stopping the alert
         frame_size: Resize frame to frame_size before sending it to the api in order to save bandwidth (H, W)
         cache_backup_period: number of minutes between each cache backup to disk
-        frame_saving_period: Send one frame over N to the api for our dataset
         cache_size: maximum number of alerts to save in cache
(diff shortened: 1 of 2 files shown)


fe51:
Hi there, thanks for the PR, 

Before diving into the review, some food for thought:

To a certain extent I'm okay if it doesn't improve our vision algos. 

By any chance, should we keep this process but only over a given period for each site? Keep this functionality (perhaps by lowering the frequency?) but, for example, only for the first X months of a new site ? Would it be relevant ?

Specifically, with regard to the temporal detection argument, wouldn't we want to obtain a series of random frames useful for our datasets ? Hence, keeping the feature saving random img but improving it to save consecutive frames ?

MateoLostanlen:
Hi @fe51, thanks for your comments. Ok I understand your point of view, so what I propose is to pass an argument to make this save optional and add the fact of saving N consecutive frames

fe51:
Yes perfect ! 

If you want, you can split it 2 features (1. making it optionnal,    2. saving N consecutives)

PR #164, opened 2023-07-27
New alerte strategy

MateoLostanlen:
Hello,

The counter we're currently using poses a problem if we have the following prediction sequence (0: no_fire, 1:fire):

1 0 1 1 0 1 1 1 1

We'll start transmitting images from frame 7 onwards, although the previous images are probably interesting.

What's more, we're not using our model's location information, so two FPs in different places will increase the counter together.

I therefore propose a new strategy, combining the results of N consecutive frames. As shown in the schema below, the idea is to perform an NMS on all the boxes amoung the N frames, then for each resulting box, sum up the score of each box with an intersection >0. In the schema, we sum up all the red boxes to obtain a score of 1.21.

The new detection threshold is conf_th * N, i.e. here 0.25 x 4 = 1. Only the red box is retained.

We then send the 4 frames used to make the prediction, in order not to lose any image.

NB: in this case, at frame 4, the old counter would be at 1, meaning that many images are lost.

image

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -159,7 +159,7 @@ class Engine:
 
     def clear_cache(self) -> None:
         """Clear local cache"""
-        for file in self._cache.rglob("*"):
+        for file in self._cache.rglob("pending*"):
             file.unlink()
 
     def _dump_cache(self) -> None:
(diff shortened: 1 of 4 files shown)


fe51 (approved):
Thanks for the peer review !

PR #165, opened 2023-07-28
Occlusion mask

MateoLostanlen:
Hi there,

The idea of this PR comes from @fe51, the idea is to use a mask of occlusion in order not to consider the predictions on a zone of the image. This will enable us to avoid FPs in the middle of the sky. This will enable us to hide specific problem areas in a quarry, for example.


fe51 (approved):
Thanks for the Peer review and the fix !

PR #166, opened 2023-07-31
limit bbox size

MateoLostanlen:
Due to a recent api problem, we propose to limit the size of bboxes sent. We send a maximum of 5 boxes with 3 decimals.

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -251,6 +251,10 @@ class Engine:
             else:
                 output_predictions = combine_predictions
 
+            # Limit bbox size in api
+            output_predictions = np.round(output_predictions, 3) # max 3 digit
+            output_predictions = output_predictions[:5, :] # max 5 bbox
+
         self._states[cam_key]["last_predictions"].append(
             (frame, preds, str(json.dumps(output_predictions.tolist())), datetime.utcnow().isoformat(), False)
         )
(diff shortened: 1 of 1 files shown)


pyroengine/engine.py
```diff
@@ -154,12 +157,16 @@ def __init__(
         self.occlusion_masks = {"-1": None}
         if isinstance(cam_creds, dict):
             for cam_id in cam_creds:
+                print()
+
```
frgfm, line 161: did you want to log that? (we won't see the print)

pyroengine/engine.py
```diff
@@ -154,12 +157,16 @@ def __init__(
...
             for cam_id in cam_creds:
+                print()
+                
                 mask_file = cache_folder + "/occlusion_masks/" + cam_id + ".jpg"
                 if os.path.isfile(mask_file):
                     self.occlusion_masks[cam_id] = cv2.imread(mask_file, 0)
                 else:
                     self.occlusion_masks[cam_id] = None
 
+        print(self.occlusion_masks)
```
frgfm, line 168: same here

frgfm (dismissed):
Thanks for the PR!
I added side comments, nothing major

MateoLostanlen:
@frgfm some commit confusion with a test I was doing can you review again

frgfm (commented):
Thanks for the edits!

fe51:
Straightforward question, I assume that the predictions are listed in descending order of probability ? (so only keeping the most relevant ones)

MateoLostanlen:
Yes it's sorted in nms function

PR #178, opened 2024-03-20
new script

MateoLostanlen:
Before updating the bastion, we add this scriipt which automatically launches an action on the pi via github in case of a problem

### DIFF
--- a/scripts/update_script.sh
+++ b/scripts/update_script.sh
@@ -6,7 +6,7 @@
 #    any others change needed
 # 
 # This script must be run with a crontab, run every hour
-# 0 * * * * bash /home/pi/pyro-engine/scripts/update_and_execute.sh >> /home/pi/pyro-engine/logfile.log 2>&1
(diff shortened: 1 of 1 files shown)


scripts/update_script.sh
```diff
@@ -1,19 +1,33 @@
...
 #- if any change:
-#    kill container
-#    rebuild docker compose
+#    pull changes
+#    any others change needed
 # 
-# This script must be run with a crontab, run every day at 3am
-# 0 3 * * * bash /home/pi/pyro-engine/scripts/update_script.sh
+# This script must be run with a crontab, run every hour
+# 0 * * * * bash /home/pi/pyro-engine/scripts/update_and_execute.sh >> /home/pi/pyro-engine/logfile.log 2>&1
```
fe51, line 9: the name of the current file does not seem to have changed so it might be 
```suggestion
# 0 * * * * bash /home/pi/pyro-engine/scripts/update_script.sh >> /home/pi/pyro-engine/logfile.log 2>&1
```
so this command for a CRON is set up at setup time, I am right ?

MateoLostanlen: oh yes thansk

PR #191, opened 2024-04-24
Add new gitAction to push docker images to dockerhub

RonanMorgan:
I have created a test env which will be launched every night so it would be good to update our image in dockerhub regularly


MateoLostanlen (approved):
Hi @RonanMorgan , thanks for your pr. Please always add a description before asking for a review. Even if it's very simple

PR #192, opened 2024-04-26
GitAction should not be launched if a PR is closed but not merged

RonanMorgan:
According to this remark : https://github.com/pyronear/pyro-api/pull/320#discussion_r1579565110


MateoLostanlen (approved):
That make sense indeed, thanks for this PR !

PR #194, opened 2024-05-20
Refactor engine

MateoLostanlen:
### Refactoring for PTZ Camera Integration

I propose a refactor of the code to prepare for the use of PTZ cameras, which is divided into two features:

**Feature 1: New ReolinkCamera**

This new `ReolinkCamera` class allows control of both PTZ cameras and the current static cameras.

**Feature 2: Parallel Processing**

In the previous version of the engine, a 30-second loop would allocate 30 seconds / (number of cameras) per camera to perform three tasks:

- Image capture
- Prediction calculation
- Processing the prediction and sending images in case of alerts

The problem was that in some cases, the image sending failed due to a lack of time. Additionally, this method does not scale well with a larger number of cameras, which will be the case with PTZ cameras that have a smaller field of view.

I propose to launch three parallel processes to manage these steps independently, optimizing the calculations as much as possible.

The code has been tested and works without any issues.


MateoLostanlen:
I will create a new PR

PR #195, opened 2024-05-20
New model

MateoLostanlen:
Following our discussion on the slack vision channel, we have a new reference model that[ has been added to huggingface](https://huggingface.co/pyronear/yolov8s). I suggest using it from now on.

This new model uses a larger 1024x1024 image instead of 384, 640. The inference on the pi is done in 3.5s instead of 1s previously, which is not a problem especially with PR #194

### DIFF
--- a/pyroengine/vision.py
+++ b/pyroengine/vision.py
@@ -15,7 +15,7 @@ from .utils import letterbox, nms, xywh2xyxy
 
 __all__ = ["Classifier"]
 
-MODEL_URL = "https://github.com/pyronear/pyro-vision/releases/download/v0.2.0/yolov8s_v001.onnx"
+MODEL_URL = "https://huggingface.co/pyronear/yolov8s/resolve/main/yolov8s.onnx"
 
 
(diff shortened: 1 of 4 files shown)


pyroengine/vision.py
```diff
@@ -11,11 +11,11 @@
 import onnxruntime
 from PIL import Image
 
-from .utils import letterbox, nms, xywh2xyxy
+from .utils import DownloadProgressBar, letterbox, nms, xywh2xyxy
 
 __all__ = ["Classifier"]
 
-MODEL_URL = "https://github.com/pyronear/pyro-vision/releases/download/v0.2.0/yolov8s_v001.onnx"
+MODEL_URL = "https://huggingface.co/pyronear/yolov8s/resolve/main/yolov8s.onnx"
```
RonanMorgan, line 18: Should not be hardcoded

RonanMorgan (changes requested):

RonanMorgan (approved):

PR #198, opened 2024-05-21
Add precommit hook

RonanMorgan:
Install precommit hook, the same configuration as in pyro-engine.


MateoLostanlen (changes requested):
Hi @RonanMorgan, thnaks for your PR ! please can you add the setup-docker-compose.sh script to the script folder and generate a docker-compose.override.yml ignore by git that will be read automatically in order to avoid changing a file track by git

MateoLostanlen (changes requested):
Hello, thanks for the update. 

Please do not change stuff that have nothing to do with this PR you should not format engine.py here

RonanMorgan:
@MateoLostanlen sorry I thought that it would be very small and there is already so much PR open so that I prefered move forward

MateoLostanlen:
@RonanMorgan what do you mean you're not going to continue with this PR?

RonanMorgan:
TODO : if we use the precommit hook, we should remove the linting part in the makefile and also clean up the workflows...

PR #200, opened 2024-05-22
Donwload model only when needed

RonanMorgan:
This code is :
- creating a meadata file with the hash of the model if this doesn't exist
-  if already existing , it compares the hash downloaded and the hash of the file
-  if the file doesn't exist or the hash are different it will download the new model

TODO : we still need to create a new path name in huggingface space in order to have a "neutral name" (like pyronear/model.onnx) and not having to change the value in the code each time @MateoLostanlen


MateoLostanlen (changes requested):
Hi Ronan, this looks togood to me. Just can you change the model path and hash based on PR #202

MateoLostanlen (approved):

PR #201, opened 2024-05-27
Add Makefile command for building code

RonanMorgan:
I was a bit lost when I tried to build the code because of the separation between the lib and the app so I have added some Makefile commands in order to better reflect this architecture and I removed one Makefile file and the requirements.txt for simplification purpose


Makefile
```diff
@@ -19,13 +19,29 @@ test:
...
+
+# update requirements.txt
+lock:
+	cd src; poetry lock
+	cd src; poetry export -f requirements.txt --without-hashes --output requirements.txt
+
 # Build the docker
-build:
+build-app:
 	docker build . -t pyronear/pyro-engine:python3.8.1-slim
```
MateoLostanlen, line 30: why do you build two images here ?

MateoLostanlen (approved):
just one small question but it LGTM

Makefile
RonanMorgan: hum build-app is building one image, with the app contained in src/ and build-lib is building the library locally.
Maybe build-image and build-lib would be a better wording ?

MateoLostanlen: Sorry, maybe it's just me, but I don't understand why we need these two lines, we can't just keep :

`docker build . -t pyronear/pyro-engine:latest`

RonanMorgan: done @MateoLostanlen

.github/workflows/docker.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build docker image
-        run: docker build . -t pyronear/pyro-engine:python3.8.1-slim
+        run: make build-app
       - name: Run docker container
-        run: docker run pyronear/pyro-engine:python3.8.1-slim python -c 'import pyroengine'
+        run: make run
+
```
RonanMorgan, line 18:
```suggestion
```

MateoLostanlen (approved):
All good thanks

PR #205, opened 2024-06-11
fix memory limit

MateoLostanlen:
Hi @RonanMorgan  , sorry I think I reviewed a bit too fast last time the pr does not work for me. I suggest you this evolution which is simpler and works without problem. Let me know if it's ok for you


RonanMorgan:
yes ok , I don't understand I am sure I tested it locally but next time I am gonna test on rpi as well

it seems that we can activate cgroups in rpi so let's keep the memory part for now I will try to activate cgroup later

RonanMorgan (approved):

PR #207, opened 2024-06-16
Fix docker

MateoLostanlen:
Quick PR:

1. network_mode: host ensure camera to be available in the container
2. we need to be sure to use the latest image with run


RonanMorgan:
note sure about network: host  ; why do you want to do it ?  which camera ?

MateoLostanlen:
hello @RonanMorgan on my pc the camera connected to the pc and used to capture images are not "pingable" without this line

RonanMorgan (approved):

PR #213, opened 2024-06-24
fix pyroclient

MateoLostanlen:
Following PR #204, the pyroclient commit was not fixed in src

I don't realy get the point of having two pyproject.toml @RonanMorgan @frgfm any thoughts ?


frgfm:
Hey there, sorry I don't get what you mean? I can only see one `pyproject.toml` 

If you mean `setup.py` and `pyproject.toml` it's because the setuptools team and the core Python teams haven't yet agreed to a unified way of doing things. It will go away (hopefully soon)

MateoLostanlen:
Hi @frgfm  no here we have pyproject.toml and src/pyproject.toml. Can't we unite them in one file?

frgfm:
My apologies! To be honest, I haven't contributed much to this repo in a while :sweat_smile: 
Back then, I think we had those separated so that the library can be used as standalone lib, but now I don't see any reason to have different pyproject and yup I agree this should be unified and similar to the API (in the sense that we need accurate version pinning, not upper/lower bound so poetry)

Happy to help if we need to make the docker image lighter & clean the build process :ok_hand:

MateoLostanlen:
Ok thanks for your answer. completely agree. I'll merge this pr to fix the bug then I'll make another to unify based on the pyro-api model

PR #216, opened 2024-07-02
Migrate to new datamodel

RonanMorgan:
This is a simplification of the data model, now the engine only need to create Detection object


MateoLostanlen:
Hello, thanks for the PR. With a first quick read I have the impression here that there is confusion between pose_id and azimuth. Cameras have preconfigured positions 0,1,2 etc ...

the api expects an azimuth right?

RonanMorgan:
fixed !

pyroengine/core.py
```diff
@@ -145,17 +142,21 @@ async def analyze_stream(self, image_queue: asyncio.Queue) -> None:
...
 
     async def night_mode(self) -> bool:
         """
-        Checks if it is nighttime for any camera.
-
-        Returns:
-            bool: True if it is daytime for all cameras, False otherwise.
+        Checks and updates the day_time attribute based on the current frame.
         """
+        try:
```
MateoLostanlen, line 153: why do you need to add this ? we are already checking all cameras just below, you will do it twice for the first one no?

MateoLostanlen (commented):
Hi @RonanMorgan, overall it's ok for me, I just need to test to validate I'll do it when we relaunch the api. I just put a small question that I did not understand

pyroengine/core.py
RonanMorgan: hum I have added that because I had a bug on these line, you want me to remove it ?

MateoLostanlen:
Sorry @RonanMorgan I mest up this PR, I have open a new one #238 for cleaner dev :)

PR #226, opened 2024-08-13
Reolink commands

MateoLostanlen:
update script with all available commands


MateoLostanlen:
test fail because of #229

PR #228, opened 2024-11-10
update test version

MateoLostanlen:
V2 is deprecated let's use v4 for unitest, let's update all versions to V4


MateoLostanlen:
unitest fail with api et's fix that in an other PR : https://github.com/pyronear/pyro-engine/issues/229

PR #241, opened 2025-02-07
Write new credentials.json format in the README

RonanMorgan:
The format of the credentials.json file has changed a bit with the new data model

PR #243, opened 2025-02-19
Send an image regularly to the API 

RonanMorgan:
The image will be used to check that everything is ok with the station

I didn't want to create a specific timer in the engine for that so I am using a random number : every loop of the engine (approx each 10sec) if the number is equal to 1 (1 chance in 100), an image is sent to the API

PS : it could take one hour to test so I have added a log, we should launch it first on a test rpi


fe51:
Hi @RonanMorgan thanks fo the PR. I think there are some commits to be pushed ? I have the feeling that the only modifications here are those from the other PR in progress #241

MateoLostanlen:
repolace by #246

PR #244, opened 2025-02-21
Fix use of input model_path in the engine

theocayla:
- Fix the use of model_path as input to instantiate the Engine
- Add checks on the model_path and onnx load


MateoLostanlen (approved):
LGTM, thanks @theocayla

PR #267, opened 2025-06-11
Update nb of consecutive frames to 8

theocayla:
The engine takes into account several consecutive frames to average the predicitons made by the model. After having tested several values of this parameters with the EvaluationPipeline (https://github.com/earthtoolsmaker/pyro-eval), we notice that setting this parameter to 8 allows to avoid several false positives (-30%) while conserving most detections (-1.5%). Tests have been performed on 70 sequences of fire and 70 sequences of false positive.
Test details are available here:

https://docs.google.com/spreadsheets/d/1v_u4cCko0vaGM2iV7nkdmvA5s_YBXFhalXDmcuKZ_6A/edit?gid=2071703334#gid=2071703334
or
[Pyro Metrics.xlsx](https://github.com/user-attachments/files/20688464/Pyro.Metrics.xlsx)

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -83,7 +83,7 @@ class Engine:
         max_bbox_size: float = 0.4,
         api_url: Optional[str] = None,
         cam_creds: Optional[Dict[str, Dict[str, str]]] = None,
-        nb_consecutive_frames: int = 4,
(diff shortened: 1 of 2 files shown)


MateoLostanlen (approved):
LGTM thanks !

PR #278, opened 2025-06-25
[WIP] Add a parameter to the Engine to control model confidence threshold

theocayla:
In the Engine we instantiate the Classifier as follows:
`self.model = Classifier(model_path=model_path, conf=0.05, max_bbox_size=max_bbox_size)`

The default value for conf is `0.15`in the Classifier definition.

The value of 0.05 is hardcoded in the call and cannot be parametrized from outside. 

This PR allows to instantiate an Engine by specifying this value.

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -79,6 +79,7 @@ class Engine:
         self,
         model_path: Optional[str] = None,
         conf_thresh: float = 0.15,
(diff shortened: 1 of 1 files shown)


MateoLostanlen:
> FIXME : fake_preds do go through post_process() hence are not filtered by confidence

@theocayla do you plan to do it here ? otherwise please open an issue

PR #312, opened 2025-11-16
Pyro camera api

MateoLostanlen:
This PR introduces a major improvement in the camera management layer. The camera API now supports every camera type through a single unified interface. Reolink PTZ cameras, RTSP streams, HTTP snapshot URL cameras and Fake cameras for development are all handled transparently using a common abstraction. This makes the system easier to configure, easier to deploy and easier to extend

The architecture is now modular. Each camera type has its own class that inherits from the same base interface. The system automatically selects the right implementation at runtime based on the camera configuration. No branching is required in the main logic. Adding a new camera type becomes trivial, since only a new subclass is needed without touching the rest of the code

Configuration has been simplified. Every camera entry in credentials.json contains the type information as backend. For example reolink, rtsp, url or fake. The engine and the camera API do not need to be changed when new deployments are added. Users only need to add the camera folder and credentials

Other improvements include a cleaner project layout. Registry logic has been centralised. Camera specific behaviours are isolated. The API routes are now consistent for all camera types. RTSP cameras ignore the pose argument without failing. PTZ cameras take the pose normally. Everything remains compatible with the existing detection and alert pipeline

This refactor was done with a strong focus on clarity, reliability and extensibility. There is no functional change for existing deployments. The behaviour is identical for Aveyron, Ardèche and Chile. Only the camera layer is more robust and future proof

### DIFF
--- a/pyro_camera_api/client/pyro_camera_api_client/client.py
+++ b/pyro_camera_api/client/pyro_camera_api_client/client.py
@@ -66,13 +66,13 @@ class PyroCameraAPIClient:
 
     def list_cameras(self) -> List[str]:
         resp = self._request("GET", "/cameras/cameras_list")
-        data = resp.json()
-        return data.get("camera_ids", [])
+        resp.raise_for_status()
+        return resp.json()
 
     def get_camera_infos(self) -> List[Dict[str, Any]]:
         resp = self._request("GET", "/cameras/camera_infos")
-        data = resp.json()
-        return data.get("cameras", [])
+        resp.raise_for_status()
+        return resp.json()
 
     # ------------------------------------------------------------------
     # Capture


README.md
```diff
@@ -19,123 +19,200 @@
...
-- `CAM_PWD`: the password to access the camera
+### Environment variables
 
-So your `.env` file should look something like this:
+Deployments usually rely on a `.env` file with information such as:
 
-```
+```text
 API_URL=https://api.pyronear.org
 CAM_USER=my_dummy_login
```
fe51: In the end, these camera credentials are only used for reolink backend/adapter right ? Hence shouldn't it be store in `credentials.json` ?

fe51 (commented):
Before going any further with the PR, here are some initial comments. I have been able to run the 2 services using "fake" camera setup (and pyro-api with envdev)

Also, great job redesigning the camera management logic! 

You'll see in the comments that I don't think the word ‘backend’ is appropriate. I've suggested adapting it and would be happy to discuss it further.

docker-compose.yml
```diff
@@ -1,16 +1,15 @@
...
+    image: pyronear/pyro-camera-api:latest
+    container_name: pyro-camera-api
     environment:
       CAM_USER: ${CAM_USER}
       CAM_PWD: ${CAM_PWD}
       MEDIAMTX_SERVER_IP: ${MEDIAMTX_SERVER_IP}
     volumes:
       - ./data:/usr/src/app/data
-    command: uvicorn main:app --host 0.0.0.0 --port 8081
     restart: always
```
fe51:
```suggestion
    restart: always
    ports:
      - "8081:8081"
```
Without the following command, I have not be able to read the api docs the outside the docker. It seems to be an issue as long as a server (outside docker services) is expected to perform requests on this API. 

Do I miss something ?

README.md
MateoLostanlen: yes why not but maybe let's change this in another pr

Issue #311, opened 2025-10-31
[Experiment] Collect 4K crop img around detection bounding box


fe51:
# Pitch
Currently, in order to save bandwidth, images sent to users (mainly firefighters) are compressed, even though they originate from 4K images.

In order to analyse alerts, users often need to zoom in on images, and detected smoke may appear pixelated at a certain zoom level.

It would be extremely useful to be able to make the best operational decisions and respond more quickly to a fire, and to have a better quality image in and around the detection area.

- #305 aims to try to see if merging 4K crop image and full image resized and compressed is relevant.
- Furthermore, it is not impossible that, on the front end, we may want to display the complete compressed image and, for example, have the 4K cropped images appear when the mouse hovers over them (see screenshot illustration below)

**We would therefore like to collect 4K cropped images around the detection area to verify that this higher-quality image helps with smoke analysis.**

_Depending on the display solution chosen, the developments made in this issue will be used to collect cropped images over the long term, to assist firefighters in their analysis of detections._

<img width="711" height="403" alt="Image" src="https://github.com/user-attachments/assets/86df4f20-3d0a-4d59-82a8-c2181933c312" />

# What should be done
- A function/class that, from 4K image captured, extract a crop image around the detection bounding box (bbow): bbox enlargement factor as a parameter (extract an image 20% larger than the bbox) 
- For each detection sent to the API, the cropped image should be send to an S3 bucket -> (all detections or only a fraction?) -> To be discussed

S3 bucket info can be provided (but for developing the feature, might be relevant to use the local S3 service from [pyro-envdev](https://github.com/pyronear/pyro-envdev))

More information to be provided (ex where to add this feature in the detection process)

Happy to discuss it ! :)

PR #320, opened 2026-01-04
Save multi resolution image (#311)

Gtortellier:
* create an image with high resolution on bboxes part and low resolution elsewhere

* save the image with the avif format and lower overall quality to save image weight


MateoLostanlen:
Hi @Gtortellier , thanks a lot for your PR. So, I've done quite a bit of testing and, although I really like the idea, unfortunately the raw approach is going to be a problem because the encoding time on Raspberry Pi is way too long and therefore not suitable for our use case. It's a shame because the camera quality with the crop was really interesting, but unfortunately we can't apply it to our use case.
That said, it's still a good idea to increase the quality at the detection level, especially for small smoke plumes in the distance. So what we've put in place (there were PRs opened for this purpose) is, in addition to the image, to send an uncompressed 4K crop around the detection. Then, on the platform, we'll display both the main image and the 4K crop in order to help better understand what the model has detected. You can see it in this PR: https://github.com/pyronear/pyro-engine/pull/359

PR #329, opened 2026-03-18
Feat/detection pose

### LINKED ISSUE #328: Adapt detection creation to pyro-api updates
Since https://github.com/pyronear/pyro-api/pull/539 on pyro-api, detection creation endpoint has been updated, and with its required parameters. 

engine need to be updated and configuration files as well (to include pose)

fe51:
The PR to close #328 _Adapt detection creation to pyro-api updates_

It introduces the following modifications : 
- Replace azimuth/azimuths with pose_ids in credentials.json, src/run.py, engine, README
- Update create_detection call: (media, azimuth, bboxes) → (media, bboxes, pose_id) to match new pyro-api endpoint
- Fix fill_empty_bboxes: use .tolist() to produce Python floats instead of np.float64
- Skip alerts with empty bboxes instead of blocking the queue -> Added but should not happen
- Add response.raise_for_status() on occlusion mask fetch to handle 404 gracefull
- Log API response body on KeyError('id') for easier debug
- Bump pinned pyroclient commit in requirements.txt to version with new create_detection signature

Now, it has to be tested on few sites, and need to work on updating credentials on all rpi fleet. 

Happy to discuss it


pyroengine/engine.py
```diff
@@ -383,7 +381,7 @@ def fill_empty_bboxes(self):
                         continue
                     filled = src.copy()
                     filled[:, -1] = 0.0  # force confidence to 0 for duplicated boxes
-                    self._alerts[i]["bboxes"] = [tuple(row) for row in filled]
+                    self._alerts[i]["bboxes"] = [tuple(row) for row in filled.tolist()]
```
MateoLostanlen: why are we updating this here ?

src/run.py
```diff
@@ -44,13 +44,11 @@ def main(args):
 
         if cam_data["type"] == "ptz":
             cam_poses = cam_data["poses"]
-            cam_azimuths = cam_data["azimuths"]
-            for pos_id, cam_azimuth in zip(cam_poses, cam_azimuths, strict=False):
-                splitted_cam_creds[ip + "_" + str(pos_id)] = (cam_data["token"], cam_azimuth, bbox_mask_url)
+            cam_pose_ids = cam_data["pose_ids"]
+            for pos_id, pose_id in zip(cam_poses, cam_pose_ids, strict=False):
```
MateoLostanlen: we need to update theses names, too similar

PR #331, opened 2026-03-25
  Extract standalone pyro_predictor package

MateoLostanlen:
The detection core (YOLO inference + temporal sliding-window state) was tightly coupled inside pyroengine, which depends on pyroclient, requests, and other
  API-related libraries. This made it impossible to use the predictor in external code (evaluation scripts, custom pipelines, CI benchmarks) without pulling
  in the full engine stack.

  Changes

  New package: pyro-predictor/

  A self-contained Python package with minimal dependencies (numpy, pillow, onnxruntime, ncnn, opencv, tqdm — no pyroclient, no requests):

  - pyro_predictor/predictor.py — Predictor class: YOLO inference + per-camera sliding-window state management (_update_states)
  - pyro_predictor/vision.py — Classifier class: NCNN/ONNX model loading and inference
  - pyro_predictor/utils.py — box_iou, nms, letterbox and helpers

  Both Predictor and Classifier accept a verbose=False parameter to suppress all log output (the package uses a named logger pyro_predictor.* instead of
  configuring the root logger, following library best practice).


Chouffe:
Any chance we can get this merged @MateoLostanlen ?

PR #337, opened 2026-04-01
Docker compose healthcheck

fe51:
Draft PR, to review after #329 

This PR introduces the following modifications : 

- engine: wrap run.py with tee to log output to engine.log; healthcheck greps for "confidence" to confirm inference is running ("confidence" is so far always in logs attesting engine is running)
- pyro-camera-api: healthcheck hits GET /health (port 8081); add curl to the Dockerfile to support it
- image versioning: replace hardcoded latest tags with ${PYRO_ENGINE_VERSION:-latest} on both services, allowing pinned deploys via .env
- docs: update README env vars section; add .env.example with all required variables documented


MateoLostanlen:
- Once "confidence" appears once, the healthcheck will pass forever, even if the engine is stuck/crashed afterward. It tests "did inference ever run" not
   "is inference currently running." A stale log from a previous run could pass the check.                                                                 
  - Night mode: The engine sleeps for 1 hour at night and skips inference entirely. During that window, if the container restarts, the healthcheck could   
  fail for 20s + 3×30s = ~110s before Docker marks it unhealthy, potentially causing unnecessary restart loops depending on restart policy.    
  
  
  
  I'm working on something else

fe51:
> Once "confidence" appears once, the healthcheck will pass forever, even if the engine is stuck/crashed afterward. It tests "did inference ever run" not
> "is inference currently running." A stale log from a previous run could pass the check.

-> not sure about thos assertion -> the aim of the command is to avoid the situation you describe by truncating log file, so it force to be a new "confidence"


> Night mode: The engine sleeps for 1 hour at night and skips inference entirely. During that window, if the container restarts, the healthcheck could
> fail for 20s + 3×30s = ~110s before Docker marks it unhealthy, potentially causing unnecessary restart loops depending on restart policy.

-> You are right here. Either improve here, either treat later.

PR #346, opened 2026-04-14
Fetch occlusion masks from pyroclient API instead of S3

MateoLostanlen:
- Replace S3-based occlusion mask fetching with pyroclient API (list_pose_masks())                       
  - Simplify cam_creds from 3-tuple (token, pose_id, bbox_mask_url) to 2-tuple (token, pose_id) across
  engine, entrypoint, and tests                                                                            
  - Parse new API response format ("mask": "(x1,y1,x2,y2)") into bbox dicts consumed by the classifier


fe51 (approved):
hi @MateoLostanlen, thank you for this quick PR. 

It looks good to me, to be quickly tested on st peray

PR #348, opened 2026-04-15
watchdog: load device-specific IPs from env file generated by Ansible

fe51:
## Pitch

- `MAIN_PI_IP` and `CAM_IPS` were hardcoded in `watchdog.py`, making the
  same script require manual edits per device
- The Pi Zero watchdog is deployed via Ansible (pi-manager-template), which
  now generates `/home/pi/watchdog.env` with the correct values per device
- `watchdog.py` reads that file at startup and overrides the two variables;
  hardcoded values remain as fallbacks so the script still works standalone

### Why no python-dotenv

The env file is generated by Ansible and intentionally kept to a trivial
`key=value` format — no interpolation, no quoted strings, no multi-line
values. Adding `python-dotenv` for two scalar overrides would be unnecessary
weight on a constrained Pi Zero. The 10-line manual parser covers the exact
format we control and nothing more.

### Env file format (generated by Ansible)

MAIN_PI_IP=192.168.1.99
CAM_IPS=192.168.1.11,192.168.1.12

### DIFF
--- a/watchdog/main_pi/watchdog.py
+++ b/watchdog/main_pi/watchdog.py
@@ -23,7 +23,27 @@ import RPi.GPIO as GPIO
 # ================= CONFIG =================
 
 RELAY_PIZERO = 16
-PIZERO_IP = "192.168.1.98"
+
+_ENV_FILE = Path("/home/pi/watchdog.env")
+
+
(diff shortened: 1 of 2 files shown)


MateoLostanlen (approved):
LGTM

PR #356, opened 2026-04-21
feed uncompressed frame to model

MateoLostanlen:
plucky-pelican v7.1.0 is compression-resistant, so the JPEG encode/decode round-trip before inference is no longer needed. The frame is now passed to the model uncompressed. JPEG bytes are still produced for API uploads and on-disk backups


fe51 (approved):
LGTM ! lets see how it will perform !

PR #357, opened 2026-04-21
cache model weights under versioned slug and purge stale ones

MateoLostanlen:
Before this change, bumping MODEL_REPO_ID (e.g. plucky-pelican_v7.0.0 → v7.1.0) had no effect on existing deployments: the local archive name (ncnn_cpu.tar.gz) and extract dir (ncnn_cpu/) 
  are unversioned, so the is_file() / is_dir() guards skipped the re-download and NCNN loaded the old weights.
                                                                                                                                                                                              
  Now cached weights live under data/models/{MODEL_SLUG}/ (slug derived from MODEL_REPO_ID). A version bump lands in a fresh path, and on download we purge any other slug plus the pre-slug  
  flat files to reclaim disk on edge devices

### DIFF
--- a/pyro-predictor/pyro_predictor/vision.py
+++ b/pyro-predictor/pyro_predictor/vision.py
@@ -6,6 +6,7 @@
 import logging
 import pathlib
 import platform
+import shutil
 import tarfile
 from typing import Tuple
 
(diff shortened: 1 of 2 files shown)


fe51 (approved):
Nice catch and quick fix ! Thanks !

PR #358, opened 2026-04-22
fix: replace single-IP ping with multi-method internet check

fe51:
# Pitch
ICMP ping to public IPs (1.1.1.1, 8.8.8.8) can be blocked on mobile/4G networks even when real internet access works. This caused false-positive internet failures on some watchdog CRON run

## Modifications

Replaces ping_host(INTERNET_IP) with internet_check_ok() which tries lightweight HTTP/HTTPS probes first (generate_204 endpoints) and falls back to pinging multiple IPs only if all HTTP checks fail. No new dependencies — uses stdlib urllib.request.

------

### Hypothesis

ICMP — Internet Control Message Protocol is a low-level network protocol used for diagnostics (ping, traceroute, "host unreachable" messages). It operates below TCP/UDP — no ports, just raw IP packets with a type field.

Why operators might block it:

- DDoS mitigation — ICMP flood attacks are trivial to launch, so many carriers simply drop inbound/outbound ICMP at their edge to protect infrastructure.
- Traffic prioritization — on constrained 4G links, operators deprioritize or discard "non-revenue" control traffic; ICMP is the easiest thing to drop without breaking real applications.
- Firewalling policy — some operators run CGNAT + stateful firewalls that only track TCP/UDP sessions; ICMP has no session concept so it gets dropped by default.
- Prevent network mapping — blocking ICMP makes it harder for outsiders to probe the operator's internal topology.

HTTP/HTTPS traffic on port 443 is virtually never blocked because that would break the entire web — which is why it's a more reliable signal for "internet is actually working."

### DIFF
--- a/watchdog/pi_zero/watchdog.py
+++ b/watchdog/pi_zero/watchdog.py
@@ -59,7 +59,8 @@ CAM_IPS: list[str] = [ip.strip() for ip in _cam_ips_raw.split(",") if ip.strip()
 _INTERNET_HTTP_URLS = [
     "https://clients3.google.com/generate_204",
     "https://connectivitycheck.gstatic.com/generate_204",
-    "https://example.com",
+    "http://cp.cloudflare.com",
+    "www.msftconnecttest.com/connecttest.txt"
 ]
 _INTERNET_PING_IPS = ["1.1.1.1", "8.8.8.8", "9.9.9.9"]
 
--- a/watchdog/pi_zero/watchdog.py
+++ b/watchdog/pi_zero/watchdog.py
@@ -56,7 +56,12 @@ CAM_IPS: list[str] = [ip.strip() for ip in _cam_ips_raw.split(",") if ip.strip()
     "192.168.1.11",
     "192.168.1.12",
 ]
-INTERNET_IP = "1.1.1.1"
+_INTERNET_HTTP_URLS = [
+    "https://clients3.google.com/generate_204",
+    "https://connectivitycheck.gstatic.com/generate_204",
+    "https://example.com",
+]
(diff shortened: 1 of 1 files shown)


watchdog/pi_zero/watchdog.py
```diff
@@ -56,7 +56,12 @@ def _load_env(path: Path) -> dict:
     "192.168.1.11",
     "192.168.1.12",
 ]
-INTERNET_IP = "1.1.1.1"
+_INTERNET_HTTP_URLS = [
+    "https://clients3.google.com/generate_204",
+    "https://connectivitycheck.gstatic.com/generate_204",
+    "https://example.com",
```
MateoLostanlen: From what I understood, "https://example.com" is not really made for this use case. quite heavy compre to the previous one? I say either delete it or update with  

http://cp.cloudflare.com
www.msftconnecttest.com/connecttest.txt

MateoLostanlen (dismissed):
Small efficiency comment but ok for me, thanks

watchdog/pi_zero/watchdog.py
fe51: What do you mean here ? The page size https://example.com of  is too big ?

Seems not that big but okay to replace by your suggestion

MateoLostanlen: with this page you will fetch whole html, bigger than others pages that are made for ping, but this is detail no wories. https://example.com/ is not really big anyway

fe51:
Fixed @MateoLostanlen , can you approve and I merge after :)  ?

MateoLostanlen:
hi @fe51 can you fix style first ?

fe51:
done @MateoLostanlen ! thanks for the catch

MateoLostanlen (approved):

PR #359, opened 2026-04-26
send 224x224 detection crop alongside full frame

MateoLostanlen:
## Summary
- Engine now uploads a tight square crop around the detection bbox in addition to the full frame, so users can confirm/dismiss smoke much faster from the dashboard without panning the full image.
- Crop is taken on the original (pre-resize) frame: 20% padding on the largest bbox dim, squared, clamped to image borders, then resized to 224×224 (JPEG q=95 if downscaling, q=100 with no chroma subsampling otherwise).
- Forwarded to `pyroclient.create_detection` via a new `crop=` kwarg.

## Depends on
- API + `pyroclient` change: https://github.com/pyronear/pyro-api/pull/575

Until that PR is merged and the lockfile is bumped, alert uploads will raise `TypeError` — do **not** merge in isolation.


fe51 (approved):
thanks @MateoLostanlen

PR #360, opened 2026-04-27
Fix reolink patrol

### LINKED ISSUE #347: Camera occasionally stuck on a single position, returning identical images
## Pitch

Occasionally, cameras get stuck on a fixed position. When requesting images for multiple different positions, the camera keeps returning the same image instead of updating to the requested viewpoints.

This issue has several impacts:

Multiple alerts are generated with misleading azimuth values
Duplicate alerts are triggered for the same fire, leading to a poor user experience
While the issue persists, the camera effectively monitors only a single position

A camera reboot resolves the problem.

## Solution suggestion 
- Monitor that the latest images for each position are actually distinct (frequency to be defined, depending on cost vs. proba of occurence of the issue)
- Automatically reboot the camera when the issue is detected
- Ensure proper logging when the error occurs

Happy to discuss it

MateoLostanlen:
The Reolink cameras regularly freeze. To avoid this issue, we propose regularly checking whether the camera poses are changing. If they are not, we reboot the camera


pyro_camera_api/pyro_camera_api/main.py
```diff
@@ -66,6 +74,22 @@ async def lifespan(app: FastAPI):
...
 
+        stuck_detector_enabled = os.getenv("ENABLE_STUCK_DETECTOR", "true").strip().lower() in (
+            "1",
+            "true",
+            "yes",
+            "on",
+        )
+        if stuck_detector_enabled and getattr(cam, "cam_type", "static") == "ptz" and hasattr(cam, "reboot_camera"):
+            stuck_flag = threading.Event()
+            stuck_thread = threading.Thread(target=stuck_check_loop, args=(cam_id, stuck_flag), daemon=True)
```
fe51, line 85: Minor naming question: `stuck_check_loop` calls its first argument `camera_ip`, but `main.py` passes `cam_id` to it (the `CAMERA_REGISTRY` loop variable). Since `CAMERA_REGISTRY` is keyed by IP here the name is technically correct, but renaming the parameter to `cam_id` in `stuck_detector.py` would align with `main.py` and `patrol.py` conventions.
what do you think ?

fe51 (commented):
Thanks a lot for the PR, seems mainly fine, but one quesiton

`stuck_check_loop` runs in a separate thread from `patrol_loop`, and both access cam.last_images concurrently. 

`stuck_check_loop` runs in a separate thread from `patrol_loop`, and both access `cam.last_images` concurrently. The `cam.last_images.clear()` call after a reboot (`stuck_detector.py` line 166) can race with the patrol loop iterating the same dict and raise `RuntimeError: dictionary changed size during iteration`. Would it be safe to replace `clear()` with resetting only the known pose keys, or add a small lock shared with the patrol loop?
=> I am not mastering threading but I spot this potentiel conflit -> What do you think ? Would it be an issue ?

MateoLostanlen:
Hi @fe51 Good catch ! The real race is in `stuck_check_loop`: it iterates `cam.last_images.items()` while the patrol loop can add new pose keys at the same time, especially right after `clear()` repopulates them. That throws `RuntimeError: dictionary changed size during iteration` and silently kills the detector thread

Fixed in 5558279: I now read by known pose key (`cam.cam_poses`, set at startup) instead of iterating the dict, and reset by key (`last_images[p] = None`) instead of calling `clear()`. No dict mutation, no lock needed

pyro_camera_api/pyro_camera_api/main.py
MateoLostanlen: Good point. I’d rather avoid renaming only stuck_detector.py in this PR: in pyro-engine, cam_id means the camera-pose identifier ("{ip}_{p}"), while here we are passing the CAMERA_REGISTRY key, which is the camera IP. So the cleaner fix would actually be to rename the cam_id loop variable in main.py to camera_ip, and possibly harmonize similar cases elsewhere. I’ll open a follow-up issue for that cleanup instead of expanding this PR scope

MateoLostanlen: https://github.com/pyronear/pyro-engine/issues/368

fe51: okay to me !

fe51:
Okay, thanks for the fix !

fe51 (approved):

PR #363, opened 2026-04-29
feat: add end-to-end relay test scripts for both Pis

MateoLostanlen:
- New watchdog/main_pi/relay_check.py and watchdog/pi_zero/relay_check.py: ping the controlled device, drive the relay LOW for POWER_OFF_TIME, verify the device drops, restore HIGH, and verify it returns. Useful after transport to confirm GPIO -> relay -> power wiring is intact.
- Pi Zero script supports --relay {main,cams,all}; cams test pings the first configured camera as a proxy for the 12V rail.
- Drop detection requires two consecutive failed pings and runs across the full cut window to avoid false positives from transient packet loss or slow-discharging loads.
- Both scripts restore relays HIGH in finally to avoid leaving a rail off on early exit.


fe51 (changes requested):
Hi @MateoLostanlen, thanks for the PR ! 

-`scripts/setup_patrol_presets.py`could you add common command to run ? (on the main pi I guess) 

- For relay_check scripts, one to be launch from ssh main pi and the other one for pi zero, what is the strat ? Access pi zero from wifi seems to be the best strat ?

Also, what are your thoughts on the relay testing protocol? 
As relay scripts are supposed to work (and have been tested multiple time); what we want to do now, producting a station, is to verify that the relay devices are functioning correctly and are properly connected under real-world conditions, and that is what this PR does.
From now on, the idea is simply to run these two scripts, isn’t it?

PR #365, opened 2026-05-01
fix: skip noon pose image when engine starts after noon

MateoLostanlen:
- Engine used to send the daily pose image immediately at startup whenever it booted after 12:00, since `last_pose_image_sent` was None and the `now >= today_noon` branch fired on the first inference call.
- Some stations reboot every night, so on next-day startup they would capture and upload a dark/empty night-time pose image — useless data.
- Anchor `last_pose_image_sent` at engine construction so the schedule only fires on a real noon crossing during the run; next noon after startup is the first send.
- Adds tests for startup-after-noon, noon crossing, and next-day noon.


fe51 (approved):
Hi @MateoLostanlen, thank you for this and for adding the relevant test cases

PR #373, opened 2026-06-05
Feat/setup presets tool

fe51:
Tools in your branch @MateoLostanlen  are very convenient, we shall consider add them to the main branch


MateoLostanlen (approved):
Agree with me haha

PR #386, opened 2026-07-21
feat(engine): send capture time as recorded_at on detection upload

### LINKED ISSUE #296: Update detections transmission to alert api to add the detection capture date as created_at parameter

MateoLostanlen:
## Motivation

Detections cached on the engine and uploaded in a burst all land at the same `created_at` on the API, which breaks time based sequence linking. pyronear/pyro-api#617 adds a `recorded_at` field carrying the on-device capture time.

## What changed

- `_process_alerts` now forwards the queued frame timestamp (UTC ISO, stamped at inference time) as `recorded_at` to `create_detection`.
- Fixed the `ts` annotation on `_stage_alert` (`int` to `str`, it always carried an ISO string) and asserted the forwarding in tests.

## Blocked on

pyro-api#617: `pyroclient.create_detection` must gain the optional `recorded_at` parameter, then the pin in `requirements-git.txt` and `uv.lock` needs a bump to that commit before merging. Until then the integration tests fail with `TypeError: unexpected keyword argument 'recorded_at'`.

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -658,7 +658,7 @@ class Engine(Predictor):
         self,
         context_crop: Optional[ContextCrop],
         cam_id: str,
-        ts: int,
+        ts: str,
         bboxes: list,
         jpeg_bytes: Optional[bytes] = None,
         crop_boxes: Optional[list] = None,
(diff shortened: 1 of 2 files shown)


fe51 (approved):
Thanks a lot for the feature and the test

PR #390, opened 2026-07-27
feat(camera): azimuth tracking and move-to-azimuth for PTZ cameras

MateoLostanlen:
- Reolink cameras cannot report their azimuth, so the platform loses orientation during live streams. This adds server-side dead reckoning: preset moves (patrol or `/goto_preset`) provide the reference from the pose/azimuth mapping, and calibrated timed pan moves update it incrementally. Linovision reads azimuth from hardware.
- The pose/azimuth mapping is fetched from the platform API at startup (`GET /api/v1/poses/` with the camera token, mapped by `patrol_id`), with retries until the API is reachable. A legacy `azimuths` list in `credentials.json` still takes precedence.
- New endpoint: `GET /control/azimuth` (current value, source, moving flag, zoom level and calibrated horizontal FOV for the cone opening).
- Preset moves now hold the per-camera lock while fire-and-forget adapters travel, so concurrent PTZ commands get a 409 instead of corrupting the estimate; untimed free moves and `/stop` interruptions invalidate it until the next preset move.
- Known limitation: the patrol loop drives cameras outside the lock; the platform must stop patrol before manual control (as today).

Note: a `POST /control/move_to_azimuth` endpoint (absolute move anchored on the closest preset pose) was initially part of this PR but was moved out to keep the scope on azimuth tracking for live sessions; it will come back in a follow-up PR.

### DIFF
--- a/pyro_camera_api/client/pyro_camera_api_client/client.py
+++ b/pyro_camera_api/client/pyro_camera_api_client/client.py
@@ -292,7 +292,8 @@ class PyroCameraAPIClient:
         """Return the camera's current real-world azimuth.
 
         Response: {"camera_ip", "azimuth_deg" (null when unknown),
-        "source" ("hardware" or "tracked"), "moving"}.
+        "source" ("hardware" or "tracked"), "moving", "zoom",
+        "h_fov_deg" (calibrated horizontal FOV at the current zoom)}.
         """
         params = {"camera_ip": camera_ip}
         resp = self._request("GET", "/control/azimuth", params=params)
--- a/pyro_camera_api/client/pyro_camera_api_client/client.py
+++ b/pyro_camera_api/client/pyro_camera_api_client/client.py
@@ -199,7 +199,11 @@ class PyroCameraAPIClient:
         if duration is not None:
             params["duration"] = duration
 
-        resp = self._request("POST", "/control/move", params=params)
+        # Preset moves block server-side while the camera travels (lock hold),
+        # and duration/degrees modes sleep for the move time: use a timeout
+        # that covers the longest server-side wait.
+        req_timeout = self.timeout if (pose_id is None and duration is None and degrees is None) else 30.0
+        resp = self._request("POST", "/control/move", params=params, timeout=req_timeout)
         return resp.json()
 
     # ------------------------------------------------------------------
@@ -207,9 +211,13 @@ class PyroCameraAPIClient:
     # ------------------------------------------------------------------
 
     def goto_preset(self, camera_ip: str, pose_id: int, speed: int = 50) -> Dict[str, Any]:
-        """Move to a configured preset pose. Returns immediately."""
+        """Move to a configured preset pose.
+
+        Blocks until the camera has settled: the server holds the per-camera
+        lock while fire-and-forget adapters (Reolink) travel to the pose.
+        """
         params = {"camera_ip": camera_ip, "pose_id": pose_id, "speed": speed}
-        resp = self._request("POST", "/control/goto_preset", params=params)
+        resp = self._request("POST", "/control/goto_preset", params=params, timeout=30.0)
         return resp.json()
 
     def start_move(self, camera_ip: str, direction: str, speed: int = 10) -> Dict[str, Any]:
--- a/pyro_camera_api/client/pyro_camera_api_client/client.py
+++ b/pyro_camera_api/client/pyro_camera_api_client/client.py
@@ -310,7 +310,8 @@ class PyroCameraAPIClient:
         return resp.json()
 
     def zoom(self, camera_ip: str, level: int) -> Dict[str, Any]:
-        resp = self._request("POST", f"/control/zoom/{camera_ip}/{level}")
+        # The server holds the per-camera lock while the zoom settles (~2-10 s).
+        resp = self._request("POST", f"/control/zoom/{camera_ip}/{level}", timeout=30.0)
         return resp.json()
 
     def reboot_camera(self, camera_ip: str) -> Dict[str, Any]:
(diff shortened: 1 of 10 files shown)


pyro_camera_api/pyro_camera_api/camera/adapters/reolink.py
```diff
@@ -54,6 +58,19 @@ def __init__(
         self.cam_azimuths = cam_azimuths if cam_azimuths is not None else []
         self.protocol = protocol
         self.focus_position = focus_position
+        # Dead-reckoned real-world azimuth; None until a preset move gives a reference.
+        self.current_azimuth: Optional[float] = None
+        # An empty azimuth list is normal at boot: the mapping is fetched from
+        # the platform API by camera.pose_azimuths. A non-empty mismatched list
+        # is a config error that silently disables tracking, hence the warning.
+        if self.cam_type == "ptz" and self.cam_azimuths and len(self.cam_poses) != len(self.cam_azimuths):
```
fe51, line 66: What about static / non PTZ cameras ? 
Non blocking Ithink for this PR, so I am opening a related issue https://github.com/pyronear/pyro-engine/issues/394

pyro_camera_api/pyro_camera_api/api/routes_control.py
```diff
@@ -772,6 +718,177 @@ def move_by_degrees(
...
+def get_camera_azimuth(camera_ip: str):
+    """Return the camera's current real-world azimuth in degrees.
+
+    ``source`` is "hardware" when the value is read back from the camera
+    (Linovision) and "tracked" when it is dead-reckoned from commanded moves
+    (Reolink). For tracked adapters ``azimuth_deg`` is null until a preset
+    move (patrol pass or /goto_preset) provides a reference, and after an
+    interrupted or untimed free move. ``moving`` reflects whether a blocking
+    PTZ command currently holds the per-camera lock.
+
```
fe51, line 859:
```suggestion
    Read ``azimuth_deg`` together with ``moving``: when ``moving`` is true the
    value is the destination of an in-flight preset move, not the camera's
    current position; when false it is the current position. A free pan reports
    null while it runs, so a non-null value under ``moving`` always means "en
    route to this azimuth".

    ``moving`` only tracks the per-camera lock, which the patrol loop does not
    take. A patrolling camera therefore reports ``moving: false`` while it is
    physically turning, and ``azimuth_deg`` follows the patrol poses with no
    in-transit signal. Callers that need a trustworthy position must stop the
    patrol first, as they already do for manual control.

```

pyro_camera_api/client/pyro_camera_api_client/client.py
```diff
@@ -280,6 +288,17 @@ def click_to_move(
         resp = self._request("POST", "/control/click_to_move", params=params, timeout=30.0)
         return resp.json()
 
+    def get_azimuth(self, camera_ip: str) -> Dict[str, Any]:
+        """Return the camera's current real-world azimuth.
+
+        Response: {"camera_ip", "azimuth_deg" (null when unknown),
+        "source" ("hardware" or "tracked"), "moving", "zoom",
+        "h_fov_deg" (calibrated horizontal FOV at the current zoom)}.
```
fe51, line 296:
```suggestion
        "h_fov_deg" (calibrated horizontal FOV at the current zoom)}.
        
        When "moving" is true, "azimuth_deg" is the destination of an in-flight
        preset move rather than the current position. "moving" only reflects the
        server-side PTZ lock, so a camera moving under patrol reports false.

```

fe51 (approved):
Nice work @MateoLostanlen, this is a clean way to give the platform an orientation on cameras that cannot report one. 

The invalidate-on-pan-start plus snapshot-before / apply-after shape is the right one for dead reckoning, `resolve_camera_azimuths` refusing partial mappings avoids a whole class of index-misalignment bugs, and extracting `_execute_degrees_move` is a good cleanup on the way.

I deliberately turned most of what I found into follow-up issues rather than change requests, so this can merge without dragging future evolutions into its scope. Nothing below is blocking. I left two docstring suggestions inline, and the rest is one point worth discussing.

## Capture route also moves the camera

The known limitation is fair and the operating contract covers the live-session case. Worth noting that the patrol loop is not the only unlocked mover: `GET /cameras/capture` with a `patrol_id` also issues a `ToPos` outside `MOVE_LOCKS` and without an active-stream check (`routes_cameras.py:122`), so a read-only-looking endpoint can move the camera mid-manual-move and resync the tracked azimuth under the running handler.

Exposure is probably null in practice, and the clean fix is not a one-liner (locking the whole capture would 409 manual PTZ for a few seconds on every engine capture), so I opened #395 rather than asking for it here. Happy to discuss this aspect.

## Documenting the `moving` + `azimuth_deg` contract

Two inline suggestions on the endpoint and client docstrings. The reason: the pair (`moving`, `azimuth_deg`) already encodes everything a caller needs, `moving: true` with a value meaning "en route to this azimuth" and `moving: false` meaning "current position", but that contract is nowhere written down. A client reading `azimuth_deg` alone and drawing a cone on the map will show the wrong orientation during a preset move without ever suspecting it had to look at `moving`. Same for the fact that `moving` only tracks the per-camera lock, so a patrolling camera reports `false` while physically turning.

## Follow-up issues opened

- #395: `/cameras/capture?patrol_id=` moves the camera outside the per-camera lock.
- #394: `/control/azimuth` returns null for static and non-PTZ cameras whose azimuth is a known constant in `credentials.json`.

## Minor, take it or leave it

In `click_to_move`, aborting the tilt axis when the pan `Stop` is rejected is the right call, no argument there. But the partial `result["moves"]` is discarded when the handler converts the exception into a 500 (`routes_control.py:299-301`), so the caller cannot tell that the pan did happen. A UI that retries the same click on error would move the camera twice. Passing the accumulated moves in the error detail, or just logging them before raising, covers it.

fe51 (changes requested):
sorry, comments in docstring to be included or discussed

pyro_camera_api/pyro_camera_api/camera/adapters/reolink.py
MateoLostanlen: nice catch, thanks for opening the issue !

pyro_camera_api/client/pyro_camera_api_client/client.py
MateoLostanlen: done

fe51 (approved):
Hi @MateoLostanlen , thanks for taking feedback into account ! 

Ready to merge !

PR #398, opened 2026-08-05
feat(camera): live stream aborts the running autofocus search

MateoLostanlen:
- The autofocus search is a minutes-long sweep that nothing could interrupt: a viewer starting a live stream during the search got a blurry, sweeping image.
- `/stream/start_stream` now cancels any running focus search on the camera and waits for the lens to be restored before starting the pipeline; `/focus/focus_finder` is refused (409) while a stream is active and aborts if a stream starts mid-search.
- Minimal extraction from #391 (same naming), so #391 can rebase on top; no engine change needed, the engine retries the hourly autofocus later on 409.

### DIFF
--- a/pyro_camera_api/pyro_camera_api/api/routes_stream.py
+++ b/pyro_camera_api/pyro_camera_api/api/routes_stream.py
@@ -32,6 +32,12 @@ from pyro_camera_api.utils.time_utils import update_command_time
 router = APIRouter()
 logger = logging.getLogger(__name__)
 
+# Serializes stream startup: without it, two concurrent start_stream calls
+# can both pass the idempotent check and spawn two pipelines for the same
+# camera. The untracked duplicate keeps streaming invisibly, so stream-aware
+# guards (focus, patrol) believe no stream is running.
+_START_STREAM_LOCK = threading.Lock()
+
 
 @router.post("/start_stream/{camera_ip}")
 def start_stream(camera_ip: str, request: Request):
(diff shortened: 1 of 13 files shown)


pyro_camera_api/pyro_camera_api/api/routes_stream.py
```diff
@@ -52,78 +59,94 @@ def start_stream(camera_ip: str, request: Request):
...
+        procs = get_processes(app)
+
+        # idempotent if already running
+        if is_pipeline_running(workers.get(camera_ip)) or is_process_running(procs.get(camera_ip)):
+            logger.info("Stream for %s already running", camera_ip)
+            return {"message": f"Stream for {camera_ip} already running"}
+
+        stopped_cam = stop_any_running_stream(app)
+
+        # Abort any running focus search on this camera and wait for it to release
```
fe51, line 72: **Ordering: a failed focus cancel tears down an unrelated stream**

`stop_any_running_stream()` runs before `cancel_focus_and_wait()`, so the irreversible
step happens before the step that can still fail. Only one stream is allowed at a time,
and the same-camera case already returned at the idempotency check above, so the stream
being stopped here always belongs to *another* camera.

Viewer switches from camera A to camera B while B is running a focus search:

1. `stop_any_running_stream` → A's pipeline is killed.
2. `cancel_focus_and_wait("B")` times out (slow camera, the focus `requests.post` has no
   timeout) → `409`.
3. A is dead, B never started. The viewer loses the picture they had, on a camera that
   has nothing to do with the focus search.

Swapping the two blocks fixes it: on timeout we return 409 having changed nothing, and A
keeps streaming. There's no ordering constraint — `cancel_focus_and_wait` only touches
`MOVE_LOCKS[camera_ip]` / `FOCUS_CANCEL_EVENTS[camera_ip]` for the target camera, which is
unrelated to the pipeline being stopped.

```suggestion
    # Abort any running focus search on this camera and wait for it to release
    # the move lock, so the viewer never sees a sweep or a lens restoration.
    # Done before stopping the current stream: on timeout we return 409 without
    # having torn down a stream that was working on another camera.
    if not cancel_focus_and_wait(camera_ip):
        raise HTTPException(
            status_code=409,
            detail=f"Focus operation still running on {camera_ip}, retry in a few seconds",
        )

    stopped_cam = stop_any_running_stream(app)
```

One follow-up if you take this: with the new order, the cancel event is set *before*
`stop_any_running_stream`, so an exception in there would leak a set event and block
autofocus forever. Moving `stopped_cam = stop_any_running_stream(app)` to the first line
inside the existing `try:` (which already clears the event in its `finally`) closes that:

```python
    if not cancel_focus_and_wait(camera_ip):
        raise HTTPException(...)

    try:
        stopped_cam = stop_any_running_stream(app)
        cfg_stream = STREAMS[camera_ip]
        ...
    finally:
        FOCUS_CANCEL_EVENTS[camera_ip].clear()
```

fe51 (commented):
Thanks a lot for this quick PR!

I've tested it in pre-production, and it correctly stops any ongoing azimuth adjustment, as expected.

I left one comment. It may be a marginal case since it probably doesn't happen very often, but the logic caught my attention.

I'd be interested to hear your thoughts on it. After that, I think we'll be good to merge it as is.

pyro_camera_api/pyro_camera_api/api/routes_stream.py
MateoLostanlen: good catch, edge case but easy fix it's added :)

fe51 (approved):
Thanks for taking feedback into account ! tested again, lets go

PR #404, opened 2026-08-18
Add CTronics camera adapter

romainantes:
Adds support for CTronics cameras using HTTP snapshots and ONVIF PTZ/Imaging controls.
It also extends RTSP stream path customization and adds dedicated documentation for camera adapters.

⚠️  **Branch "docs/camera-adapter-guide" must be merged before this branch.**

Changes:
New Adapter:
- Introduced CTronicsCamera supporting authenticated HTTP snapshots, ONVIF PTZ, CGI relative focus, and ONVIF Imaging.
- Registry & Config: Registered the ctronics adapter in registry.py and updated core/config.py to allow custom rtsp_path overrides.
- Dependencies: Added onvif-zeep dependency to pyproject.toml and requirements.txt.
- Documentation: Updated README.md and camera/adapters/README.md to document adapter responsibilities and configuration details.
- Tests: Added unit and real-hardware integration tests in test_ctronics.py.


MateoLostanlen (commented):
Review of the CTronics adapter. 8 correctness issues, 1 dead code block, 1 efficiency note. Details inline.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+    def set_auto_focus(self, disable: bool) -> None:
+        self._ensure_onvif()
+        if self._imaging_service is None:
+            raise RuntimeError("ONVIF Imaging service is unavailable")
+        request = self._imaging_service.create_type("SetImagingSettings")
+        request.VideoSourceToken = self._profile.VideoSourceConfiguration.SourceToken
+        request.ImagingSettings = {"Focus": {"AutoFocusMode": "MANUAL" if disable else "AUTO"}}
+        self._imaging_service.SetImagingSettings(request)
+
+    def start_zoom_focus(self, position: int) -> None:
```
MateoLostanlen, line 366: `start_zoom_focus` calls `set_manual_focus`, so it moves the focus motor, not zoom. `routes_stream.py` calls `start_zoom_focus(position=0)` after every stream stop, which sets absolute focus to 0 and stores `focus_position = 0`, and the patrol loop then reapplies it every cycle. The camera ends up permanently defocused. In reolink and linovision this method drives zoom.

pyro_camera_api/pyro_camera_api/camera/registry.py
```diff
@@ -156,6 +159,7 @@ def build_camera_object(key: str, conf: dict) -> Optional[BaseCamera]:
         snapshot_url = conf.get("url")
         if not snapshot_url:
             logger.error("Camera %s declared as URL adapter but missing 'url'", key)
+            focus_auth = (conf.get("focus_auth", "digest"),)
```
MateoLostanlen: This line landed in the wrong branch: it sits in the url adapter's error path, is never read, and builds a 1-tuple. It also fails ruff F841. The ctronics constructor call below never receives `focus_auth`, so the config key is silently ignored. Delete this line and pass `focus_auth=conf.get("focus_auth", "digest")` to `CTronicsCamera`.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        self,
+        save_images: bool = False,
+        retry_depth: int = 0,
+        should_abort: Optional[Callable[[], bool]] = None,
+    ) -> int:
+        _ = retry_depth
+        if self.cam_type == "static":
+            return self.focus_position or 0
+        initial = self.focus_position if self.focus_position is not None else (self.focus_min + self.focus_max) // 2
+        candidates = range(max(self.focus_min, initial - 50), min(self.focus_max, initial + 50) + 1, 10)
```
MateoLostanlen: `set_manual_focus` stores `focus_position` unclamped while the route accepts any int, so `position=2000` with `focus_max=1000` makes the candidate range empty and `max(scores)` raises ValueError (500 on `/focus/focus_finder`). Also, the fixed ±50 window around a guessed seed of 500 cannot recover a badly defocused camera; consider seeding from `get_focus_level()` and clamping the stored value.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+            request = self._ptz_service.create_type("GotoPreset")
+            request.ProfileToken = self.onvif_profile_token
+            request.PresetToken = self._preset_token(int(idx))
+            logger.info(
+                "CTronics GotoPreset sending: camera=%s profile=%s token=%s",
+                self.ip_address,
+                request.ProfileToken,
+                request.PresetToken,
+            )
+            response = self._ptz_service.GotoPreset(request)
```
MateoLostanlen, line 218: ONVIF `GotoPreset` returns before the head arrives, but `preset_move_hold_s` stays at the mixin default 0.0, so `MOVE_LOCKS` is released while the camera is still traveling and `/azimuth` reports the destination with `moving: false`. Add `preset_move_hold_s = 5.0` to the class body like reolink does.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        # )
+        for preset in presets:
+            if str(getattr(preset, "token", "")) == str(preset_id):
+                logger.info(
+                    "CTronics preset id=%s matched ONVIF token=%s directly",
+                    preset_id,
+                    preset.token,
+                )
+                return str(preset.token)
+        if 0 <= preset_id < len(presets):
```
MateoLostanlen: When no ONVIF token matches, the fallback resolves by list index, but `_sync_azimuth_from_pose(idx)` still records the azimuth configured for that pose id. On cameras with vendor-assigned or 1-based tokens, the camera drives to one preset while the azimuth of another is recorded, so detections get geolocated on the wrong bearing with only an INFO log. Prefer failing loudly when the token does not match.

pyro_camera_api/pyro_camera_api/core/config.py
```diff
@@ -71,7 +71,8 @@ def normalize_stream_name(name: str) -> str:
             path = cfg.get("rtsp_path", f"/Streaming/Channels/{channel}")
             input_url = f"rtsp://{USER_ENC}:{PWD_ENC}@{ip}:554{path}"
         else:
-            input_url = f"rtsp://{USER_ENC}:{PWD_ENC}@{ip}:554/h264Preview_01_sub"
+            path = cfg.get("rtsp_path", "/h264Preview_01_sub")
```
MateoLostanlen: Two gaps for a ctronics camera: `rtsp_path` is spliced with no leading slash normalization (`"rtsp_path": "11"` yields `...@ip:55411`), and the URL still uses global `CAM_USER`/`CAM_PWD` while the README documents per-camera credentials, so capture works but streaming gets a 401. A `lstrip("/")` and a per-camera `cfg.get("username", CAM_USER)` would fix both branches.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        self.port = port
+        self.protocol = protocol
+        self.snapshot_path = snapshot_path
+        self.snapshot_command = snapshot_command
+        self.timeout = timeout
+        self.model = model
+        self.cam_poses = cam_poses if cam_poses is not None else []
+        self.cam_azimuths = cam_azimuths if cam_azimuths is not None else []
+        self.onvif_port = onvif_port
+        self.onvif_protocol = onvif_protocol
```
MateoLostanlen: `onvif_protocol` is accepted from config and stored but never used: `ONVIFCamera` takes no scheme, so traffic stays plain HTTP even with `"https"` configured. Same for `self.model`, stored and never read. Either plumb them through or drop the params so config that does nothing is rejected.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        request.Focus = {
+            "Absolute": {
+                "Position": self._clamp((position - self.focus_min) / max(1, self.focus_max - self.focus_min), 0.0, 1.0)
+            }
+        }
+        return request
+
+    def set_manual_focus(self, position: int) -> None:
+        self._ensure_onvif()
+        self._imaging_service.Move(self._focus_request(position))
```
MateoLostanlen: `_ensure_onvif` raises RuntimeError, and `routes_focus.py` calls `get_focus_level`/`set_manual_focus` with no try/except, so a missing onvif-zeep or firewalled port surfaces as a stack trace 500. Also, when the imaging service is unavailable, the attribute lookup on `self._imaging_service.Move` raises AttributeError before the guard in `_focus_request` can fire. And `get_ptz_preset` returns raw zeep objects straight into the JSON response, unlike reolink's plain dicts.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        self.focus_position = int(position)
+
+    def move_focus(self, action: str, speed: Optional[int] = None) -> bool:
```
MateoLostanlen: The Hi3510 CGI focus block (`move_focus`, `focus_plus`, `focus_minus`, `stop_focus`, the digest retry and the `focus_path`/`focus_step`/`focus_speed`/`focus_auth` params) has no caller outside its own tests: every route goes through the ONVIF Imaging path. Keep one mechanism: if ONVIF absolute focus works on the tested unit, delete the CGI half; otherwise wire `set_manual_focus` to the CGI and drop the ONVIF Move half.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
```diff
@@ -0,0 +1,459 @@
...
+        request = self._ptz_service.create_type("ContinuousMove")
+        request.ProfileToken = self.onvif_profile_token
+        request.Velocity = {
+            "PanTilt": {"x": pan * move_speed, "y": tilt * move_speed},
+            "Zoom": {"x": zoom * move_speed},
+        }
+
+        self._ptz_service.ContinuousMove(request)
+
+    def _preset_token(self, preset_id: int) -> str:
```
MateoLostanlen: `_preset_token` refetches the full preset table on every ToPos, which for an 8-pose patrol adds roughly 26k SOAP calls a day; cache the token map in `_ensure_onvif` and invalidate it in `set_ptz_preset`. Also drop the per-frame INFO in `capture()` and the three INFO lines per GotoPreset to debug level, this is the hot path on a Pi writing to SD.

MateoLostanlen:
Hi @romainantes,

Thanks a lot for the PR, and sorry for the late review! I left a few inline comments. I don't have access to the camera so it's not ideal for reviewing, but let me know what you think.

Quick question by the way: why go through ONVIF rather than the camera's native API? I've heard ONVIF is often less well maintained on the firmware side.

And following on from that, if it's pure ONVIF, I'm wondering if we should have a generic ONVIF adapter directly (reusable for other brands) rather than a CTronics-specific one. What do you think?

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for pointing this out. I fixed the CTronics adapter so `start_zoom_focus()` controls zoom through ONVIF `AbsoluteMove` instead of calling `set_manual_focus()`.

`routes_stream.py` remains unchanged: stopping a stream still resets the zoom to position `0`, but it no longer changes `focus_position` or moves the focus motor. I also added a unit test verifying that the zoom command leaves the focus position untouched. The behavior was validated with the ONVIF mock; the low-end camera used for testing does not support absolute zoom moves (so it must be tested with a better one later).

pyro_camera_api/pyro_camera_api/camera/registry.py
romainantes: Thanks for catching this. I removed the misplaced `focus_auth` assignment from the URL adapter’s error path and passed `focus_auth=conf.get("focus_auth", "digest")` to `CTronicsCamera`, so the configuration value is now correctly propagated.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for pointing this out. I fixed set_manual_focus() so the requested position is clamped to the configured [focus_min, focus_max] range before it is sent to the camera and stored in focus_position.

I also updated focus_finder() to use the current focus reported by get_focus_level() as its search seed, with a fallback to the cached position or the midpoint of the configured range. The candidate window is now guaranteed to contain at least one valid position.

Added regression tests covering out-of-range focus positions and focus-finder seeding from the current camera focus level.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for catching this. I added preset_move_hold_s = 5.0 to CTronicsCamera, matching the Reolink adapter.

Since ONVIF GotoPreset returns before the camera has finished moving, this keeps the movement lock active during the expected travel time. As a result, concurrent PTZ operations are blocked appropriately and the camera is not reported as idle immediately after the preset command. I also added a regression assertion covering the configured hold duration.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for pointing this out. I removed the list-index fallback from _preset_token so preset IDs must now match an ONVIF preset token exactly.

When no matching token is found, the adapter raises ValueError before calling GotoPreset or updating the tracked azimuth. This prevents the camera from moving to one preset while the API records the azimuth of another. I also added a regression test covering vendor-assigned, non-sequential preset tokens.

pyro_camera_api/pyro_camera_api/core/config.py
romainantes: Thanks for pointing this out. I centralized RTSP input URL construction and fixed both issues:

- RTSP paths are now normalized so both "11" and "/11" produce a valid URL with exactly one slash.

- Per-camera username and password values are now used, with CAM_USER and CAM_PWD kept as fallbacks.

This applies to both CTronics and Linovision streams. I also added regression tests covering path normalization, URL encoding, per-camera credentials, and the global credential fallback.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for pointing this out. I removed both unused configuration options from the CTronics adapter (used at first for testing):

- onvif_protocol is no longer accepted because the current ONVIFCamera integration does not provide a scheme configuration and would still use HTTP.

- model was not consumed by the CTronics adapter or any related logic.

Both options were also removed from the registry wiring and documentation. I added regression tests confirming that unsupported model and onvif_protocol arguments are rejected instead of being silently ignored.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: Thanks for pointing this out. I addressed all three issues:

- ONVIF initialization failures are now normalized as RuntimeError, and the manual-focus and focus-status routes translate them into controlled 503 Service Unavailable responses.

- set_manual_focus now explicitly checks that the Imaging service is available before calling Move.

- CTronics presets are converted to plain dictionaries containing token and name, so they can be serialized safely in API responses.

I also added regression tests covering ONVIF initialization failures, unavailable Imaging services, focus route error handling, and JSON-safe preset serialization.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: I have validated the ONVIF absolute-focus implementation on my CTronics camera using two distinct positions (100 and 900). The focus changed visibly, and GetStatus returned the corresponding absolute values.

Based on that validation, I have kept ONVIF Imaging as the single focus mechanism and removed the unused Hi3510 CGI implementation, including move_focus, focus_plus, focus_minus, stop_focus, the Digest retry logic, and the focus_path/focus_step/focus_speed/focus_auth parameters (used at first for validations).

The existing focus tests were migrated to cover ONVIF absolute movement and status reporting, and the hardware test now verifies both positions directly. This avoids maintaining two competing focus implementations and keeps the adapter aligned with the focus routes already using ONVIF.

pyro_camera_api/pyro_camera_api/camera/adapters/ctronics.py
romainantes: I added a preset-token cache initialized during ONVIF setup, so repeated ToPos calls no longer fetch the complete preset table every time. The cache is invalidated after SetPreset and rebuilt on the next preset access to keep it consistent.

I also lowered the per-frame capture log and all GotoPreset logs from INFO to DEBUG, reducing SD-card writes on Raspberry Pi deployments.

Regression tests now cover preset-cache reuse, cache invalidation, and unmatched preset tokens.

romainantes:
> Hi @romainantes,
> 
> Thanks a lot for the PR, and sorry for the late review! I left a few inline comments. I don't have access to the camera so it's not ideal for reviewing, but let me know what you think.
> 
> Quick question by the way: why go through ONVIF rather than the camera's native API? I've heard ONVIF is often less well maintained on the firmware side.
> 
> And following on from that, if it's pure ONVIF, I'm wondering if we should have a generic ONVIF adapter directly (reusable for other brands) rather than a CTronics-specific one. What do you think?

Thanks for the review and for raising this!

I chose ONVIF because the camera’s native API did not expose a sufficiently complete and reliable interface for all the operations required by the adapter. In particular, I had difficulty covering PTZ presets, absolute focus, focus status, and the different control paths consistently through the native API. ONVIF provided a single control interface for these operations, and I validated the relevant absolute-focus behavior on the available CTronics camera hardware.

That said, I am aware that ONVIF support can vary between firmware versions and vendors. The adapter therefore still handles several CTronics-specific details, such as snapshot capture, connection defaults, profile selection, preset-token behavior, and the camera’s advertised Imaging/PTZ capabilities.

I agree that a generic ONVIF adapter could be valuable in the future, especially for the PTZ and Imaging functionality. However, I would prefer to keep this PR CTronics-specific for now because the current implementation still contains vendor-specific behavior and has only been validated on CTronics hardware.

MateoLostanlen:
Hi @romainantes

Thanks for all the updates, that was a lot of work and the adapter is in much better shape now.

I went through the ten points and checked the code, not just the replies. All of them are addressed: zoom no longer touches the focus motor, the preset index fallback is gone and now fails loudly, focus is clamped and seeded from the camera, ONVIF errors return 503, the CGI half is deleted, and the preset cache plus DEBUG logs are in. Camera API tests pass locally (73 passed, 7 skipped).

Three things before this can merge:

1. The branch needs a rebase on `develop`, there are conflicts on both READMEs since #407 landed.
2. `ruff format --check` fails on `test_config.py`, `test_ctronics.py` and `test_focus_cancel.py`. A `make style` should do it.
3. No CI check has run on the PR at all, probably because of the conflicts. Worth confirming green after the rebase.

A few small things I noticed while reading:

- In `_ensure_onvif`, `self._ptz_service` is assigned before `_refresh_preset_cache()`. If `GetPresets` fails once, the init raises but `_ptz_service` stays set, so the next call returns early and `_imaging_service` is never created. Focus then stays broken for good after a single transient failure. Assigning `_ptz_service` last, or resetting the attributes in the `except`, would fix it.
- `USER_ENC` / `PWD_ENC` in `core/config.py` are unused now after the RTSP refactor.
- `README.md` still lists `model` as a CTronics option.
- `focus_finder` seeds correctly now, but the window is still ±50 in steps of 10 over a 0 to 1000 range, so a badly defocused camera still will not be recovered. Fine to leave as is, just noting it.

Also worth flagging: `tests.yml` only runs `pytest tests/`, so `pyro_camera_api/tests/` never runs on CI. That is a repo-wide gap rather than something for this PR, but it means your new tests are not actually gating anything yet.

On the ONVIF question, your reasoning makes sense and I am fine keeping this CTronics specific. Just noting that after removing the CGI path, `capture()` is about the only vendor specific piece left, so splitting a generic ONVIF base out of this will be easy the day a second ONVIF camera shows up. No need to do it here.

Last one: you mentioned absolute zoom could not be validated on your unit. Can we get it tested on a camera that supports it before we rely on it in the stream stop path?

romainantes:
> Hi @romainantes
> 
> Thanks for all the updates, that was a lot of work and the adapter is in much better shape now.
> 
> I went through the ten points and checked the code, not just the replies. All of them are addressed: zoom no longer touches the focus motor, the preset index fallback is gone and now fails loudly, focus is clamped and seeded from the camera, ONVIF errors return 503, the CGI half is deleted, and the preset cache plus DEBUG logs are in. Camera API tests pass locally (73 passed, 7 skipped).
> 
> Three things before this can merge:
> 
> 1. The branch needs a rebase on `develop`, there are conflicts on both READMEs since [Docs/camera adapter guide #407](https://github.com/pyronear/pyro-engine/pull/407) landed.
> 2. `ruff format --check` fails on `test_config.py`, `test_ctronics.py` and `test_focus_cancel.py`. A `make style` should do it.
> 3. No CI check has run on the PR at all, probably because of the conflicts. Worth confirming green after the rebase.
> 
> A few small things I noticed while reading:
> 
> * In `_ensure_onvif`, `self._ptz_service` is assigned before `_refresh_preset_cache()`. If `GetPresets` fails once, the init raises but `_ptz_service` stays set, so the next call returns early and `_imaging_service` is never created. Focus then stays broken for good after a single transient failure. Assigning `_ptz_service` last, or resetting the attributes in the `except`, would fix it.
> * `USER_ENC` / `PWD_ENC` in `core/config.py` are unused now after the RTSP refactor.
> * `README.md` still lists `model` as a CTronics option.
> * `focus_finder` seeds correctly now, but the window is still ±50 in steps of 10 over a 0 to 1000 range, so a badly defocused camera still will not be recovered. Fine to leave as is, just noting it.
> 
> Also worth flagging: `tests.yml` only runs `pytest tests/`, so `pyro_camera_api/tests/` never runs on CI. That is a repo-wide gap rather than something for this PR, but it means your new tests are not actually gating anything yet.
> 
> On the ONVIF question, your reasoning makes sense and I am fine keeping this CTronics specific. Just noting that after removing the CGI path, `capture()` is about the only vendor specific piece left, so splitting a generic ONVIF base out of this will be easy the day a second ONVIF camera shows up. No need to do it here.
> 
> Last one: you mentioned absolute zoom could not be validated on your unit. Can we get it tested on a camera that supports it before we rely on it in the stream stop path?

1. The branch needs a rebase on develop, there are conflicts on both READMEs since Docs/camera adapter guide #407 landed.
=> I wasn't able to perform the rebase correctly, so I'd appreciate it if you could take care of it.
2. ruff format --check fails on test_config.py, test_ctronics.py and test_focus_cancel.py. A make style should do it.
=> I've fixed the formatting issues reported by Ruff on all three files.
3. No CI check has run on the PR at all, probably because of the conflicts. Worth confirming green after the rebase.
=> Agreed. Once the rebase is done, I'll check the CI and address any issues if needed.

"In _ensure_onvif, self._ptz_service is assigned before _refresh_preset_cache(). If GetPresets fails once, the init raises but _ptz_service stays set, so the next call returns early and _imaging_service is never created. Focus then stays broken for good after a single transient failure. Assigning _ptz_service last, or resetting the attributes in the except, would fix it."
=> This was a real retry issue: _ptz_service could remain set after _refresh_preset_cache() failed, causing subsequent calls to skip ONVIF initialization. I reset the ONVIF state in the exception path and added a regression test covering a transient GetPresets failure followed by a successful retry.
USER_ENC and PWD_ENC were indeed unused after the RTSP refactor, so I removed them. The CTronics README entry has been updated to remove the unsupported model option. I left the focus_finder search window unchanged for now, as the current behavior is correct and the wider recovery range can be addressed separately if needed.

"Also worth flagging: tests.yml only runs pytest tests/, so pyro_camera_api/tests/ never runs on CI. That is a repo-wide gap rather than something for this PR, but it means your new tests are not actually gating anything yet."
=> I hadn't noticed that because I was running the tests manually, but you're right. This is definitely something we should address separately by opening a story to update tests.yml.

"On the ONVIF question, your reasoning makes sense and I am fine keeping this CTronics specific. Just noting that after removing the CGI path, capture() is about the only vendor specific piece left, so splitting a generic ONVIF base out of this will be easy the day a second ONVIF camera shows up. No need to do it here."
=> Regarding ONVIF, that makes sense. We can open a separate story to introduce a generic ONVIF base and move the common logic there.

"Last one: you mentioned absolute zoom could not be validated on your unit. Can we get it tested on a camera that supports it before we rely on it in the stream stop path?"
=> My current camera does not support ONVIF absolute zoom, so I could not validate the physical movement on that unit. The existing hardware test already covers start_zoom_focus; it needs to be run against a camera that supports ONVIF AbsoluteMove zoom before relying on it operationally. The stream stop path remains best-effort and safely ignores failures when the camera does not support this capability.

PR #405, opened 2026-08-18
fix(engine): send empty bboxes and unblock the alert upload queue

### LINKED ISSUE #402: Alerts stuck in cache: placeholder bbox rejected by the API (KeyError 'id')
On a station, alerts fail to upload and stay stuck in cache:

```
INFO: Camera '192.168.1.12_0' - Sending alert from 2026-08-17T17:01:59.616844+00:00...
ERROR: Camera '192.168.1.12_0' - unable to upload cache
ERROR: 'id'
```

`'id'` is the KeyError raised by `response.json()["id"]` in `Engine._process_alerts`: the API answered with an error body instead of a detection.

Cause: `fill_empty_bboxes` stamps `PLACEHOLDER_BBOX = (0.0, 0.0, 0.0001, 0.0001, 0.0)` on frames with no detection. pyroclient serializes coords with 3 decimals, so it becomes `(0,0,0.000,0.000,0)` and the API rejects it with a 422 (`xmin & ymin are expected to be respectively smaller than xmax & ymax`). Since pyro-api #624, a frame with no detection must be sent with an **empty** bbox list, which the API attaches to ongoing sequences (or answers 204).

Two problems:

1. The placeholder bbox is obsolete and now systematically rejected.
2. The failure is not observable and blocks the queue: the status code is never checked, the response body is never logged, and the `break` keeps retrying the same non-retryable alert forever (above, an alert from 17:01 was still being retried at 18:05), so every alert behind it is stuck too.

Suggested fix:

- drop `PLACEHOLDER_BBOX` / `fill_empty_bboxes` and send `bboxes=[]` as is
- check `response.status_code` instead of `json()["id"]`: log status + body, drop the alert on 4xx (a retry can only fail the same way), keep it cached on 5xx / connection errors

MateoLostanlen:
Fixes #402.

- Frames with no detection were stamped with a placeholder bbox `(0, 0, 0.0001, 0.0001, 0)`. pyroclient serializes coordinates with 3 decimals, so `xmin == xmax` and the API rejects the alert with a 422. Since pyro-api #624 those frames are reported with an empty bbox list, which the API attaches to the ongoing sequences of the pose (or answers 204).
- The upload result was read as `response.json()["id"]`, so a rejection surfaced as a bare `KeyError` with no status code and no body, and the `break` replayed the same doomed alert forever, stalling every alert queued behind it. The status code is now checked: retry on 5xx and on 408/425/429, drop and log status plus body on any other non-2xx, accept 204 without a body.
- A `ValueError` raised by the client while serializing the payload is final too, so the alert is dropped rather than retried indefinitely.
- A 2xx whose body carries no detection id keeps the alert cached: the detection was not stored.

### DIFF
--- a/pyroengine/engine.py
+++ b/pyroengine/engine.py
@@ -25,6 +25,9 @@ from requests.models import Response
 
 __all__ = ["ContextCrop", "Engine"]
 
+# Client errors worth retrying later; every other 4xx is a permanent rejection of the payload.
+RETRYABLE_STATUS = frozenset({408, 425, 429})
+
 logging.basicConfig(format="%(asctime)s | %(levelname)s: %(message)s", level=logging.INFO, force=True)
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


fe51 (approved):
Thank you for the quick fix @MateoLostanlen

PR #406, opened 2026-08-18
fix(camera-api): Linovision PTZ click_to_move and zoom control

MateoLostanlen:
click_to_move did nothing on Linovision (no calibrated speed tables, only Reolink has them) and zoom commands were broken by wrong ISAPI units. Field-validated on a deployed station.

- click_to_move now uses hardware relative moves on Linovision (ISAPI absolute positioning with position readback) instead of timed moves. The FOV at zoom is derived optically from the datasheet wide end (55x33 deg, 25x), with exact pinhole projection of the click. Other uncalibrated adapters fall back to a rough speed proxy, matching /move.
- Fixes the ISAPI zoom units: the status reports absoluteZoom in tenths (10 = 1.0x) while the absoluteEx command takes the ratio directly. The 0-64 API zoom level maps linearly to the 1-25x optical range, and the post-zoom settle lock is sized from the hardware zoom readback instead of a fixed 10 s worst case.
- The QR-based FOV measurement method is documented in tools/README.md in case mid-zoom clicks ever need verification.


fe51 (approved):
Hi @MateoLostanlen, thank you for the quick PR

Tested and works as expected. 

Just noticed, for another PR : pyro_camera_api/pyro_camera_api/api/routes_control.py is starting to be really big, might be relevant a refacto some day.

PR #407, opened 2026-08-18
Docs/camera adapter guide

fe51:
Docs only, no code changes.

## Why

The camera API has six adapters and a capability model, none of it written down. Anyone plugging in a new camera today has to reconstruct the design from `registry.py`, `base.py` and the route
handlers. Two questions have no answer in the repo: *do I even need a new adapter* (the `rtsp`,
`url` and `rest` adapters cover a lot with config alone), and *what exactly must my class
implement*.

## What it adds

One guide at `pyro_camera_api/pyro_camera_api/camera/adapters/README.md`: what an adapter is for,
a table of the six existing ones to check before writing code, the `capture()` contract, the
capabilities and how far to go, a commented skeleton, and the pitfalls. The root `README.md`
links to it and gains the missing `rest` adapter in its list.

It lives in the adapters directory so GitHub renders it when browsing the folder, which is where
you land when you go look at existing adapters. Same convention as `tools/` and `setup_presets/`.

## Worth a look later

Documented as-is, not changed here:

- **An implicit contract sits next to the explicit one.** Six methods are resolved by `hasattr()`
  in the routes and declared nowhere in `base.py` (`set_auto_focus`, `focus_finder`,
  `start_zoom_focus`, `get_ptz_preset`, `set_ptz_preset`, `reboot_camera`). An adapter can
  implement both mixins fully, pass mypy, and still 400 on six endpoints. Worst case is silent:
  without `reboot_camera`, `main.py` never starts the stuck detector and logs nothing.
- **Streaming is outside the adapter layer.** Stream URLs are built in `core/config.py` from the
  IP, assuming RTSP on 554 with a Reolink or Linovision path. A camera off that convention
  captures fine and cannot stream.
- **`adapter: "reolink"` degrades silently.** Speed tables are keyed by exact model, so the
  generic value falls back to the 823S2 calibration with only a warning.


MateoLostanlen:
Checked every claim in the guide against the code. Docs only, CI green, nothing risky. A few things are inaccurate enough to mislead someone writing an adapter.

**`focus_finder()` is not `hasattr()`-guarded.** The optional-methods table lists it under the methods routes look up with `hasattr()` and degrade gracefully. `routes_focus.py:122` only checks `isinstance(cam, FocusMixin)`, then calls `cam.focus_finder(...)` at line 149 unguarded. A `FocusMixin` adapter without `focus_finder` raises `AttributeError` and returns 500, not a clean 400. Same correction applies to "still 400 on six endpoints" in the PR description.

**The `rest` row omits `response`.** The table lists `url`, `headers`, `json_path`, `encoding`, but `response` (`rest.py:50`, default `"image"`) is the switch that enables the JSON path. Left out, `json_path` and `encoding` are ignored and the capture silently fails.

**The `url` adapter dispatches on the literal string `CGIProxy.fcgi`** (`url.py:161`), not on the presence of `usr`/`pwd`. "It exposes a snapshot URL returning an image directly, use `url`" is a trap: a credential free public snapshot URL falls into the digest branch and returns `None` with "requires inline credentials". Worth one sentence.

**`focus_position` is applied once per patrol cycle**, after the return to pose 0 (`patrol.py:115-124`), not on every pass over a preset.

Nits:

* `/control/reboot` returns 501, not 400 (`routes_control.py:973`). Real paths are `/control/reboot/{camera_ip}` and `/control/zoom/{camera_ip}/{level}`.
* "Required configuration" overstates: `ip_address` falls back to the registry key, `poses`/`azimuths` default to `[]`. Only `rtsp_url` and `url` actually make `build_camera_object` return `None`.
* Step 3 "export in `__init__.py`": the registry imports the modules directly and `__init__.py` currently exports 4 of 6 (no Linovision, no Mock). Convention rather than wiring.

Verified correct: the `base.py` contract and mixin semantics, isinstance to 400 on the mixin routes, import time registry build, the reolink speed table fallback with warning (`routes_control.py:72`), stream URLs built in `core/config.py` from the IP on 554, the stuck detector silently skipped without `reboot_camera` (`main.py:84`), both `reboot_camera` snippets, the Linovision `set_auto_focus` stub behaviour, and the patrol `move_camera("ToPos", idx=pose, speed=50)` loop.

Fix the `focus_finder` row and the `response` field, the rest is polish. The "is my camera already supported" table is the part that saves real work.

fe51:
@MateoLostanlen thks for the feedbacks, I have taken them into account and try to simplify the docs.

MateoLostanlen:
Second pass on the updated version. The August points are all fixed, thanks. Three groups of remaining issues, the first one is what I'd like addressed before merge.

**1. PTZ configuration and the skeleton**

- The skeleton's `get_azimuth()` returns `self.current_azimuth`, but its `__init__` never sets it. `azimuth_source` defaults to `"tracked"` and `routes_control.py` calls `get_azimuth()` unguarded, so an adapter copied as is raises `AttributeError` (500) on the first `GET /control/azimuth` instead of the `None` the inline comment promises. Add `self.current_azimuth = None` to the skeleton.
- The table's configuration column omits `type: "ptz"`. `registry.py:98` defaults to `"static"` and `main.py:67` only starts the patrol loop for `"ptz"`, so a PTZ camera without that field registers fine and never patrols. That's the first field to get wrong.
- The table also omits plain `reolink`, which the root README, both JSON examples and this doc's own diagram use (`registry.py:102` matches by substring). The Pitfalls bullet describes the fallback the table denies exists; one `reolink` row pointing at Pitfalls reconciles them.

**2. Worker behaviour and capture failures**

- The stuck detector paragraph is off: `main.py:84` also requires `ENABLE_STUCK_DETECTOR` (not mentioned) and `hasattr(cam, "reboot_camera")`, and the missing method case logs nothing, the detector just never starts. The 501 in the table row only applies to `POST /control/reboot`. Also, a patrol loop with no poses exits after one warning rather than "failing continuously".
- "reads `None` as no image this time" understates the static loop: `rtsp`, `url` and `rest` are all forced static, and two consecutive `None` captures trigger a 30 minute skip window (`patrol.py`, `MAX_FAILS_BEFORE_SKIP = 2`). Worth a sentence in the contract section so adapter authors don't return `None` casually on transient errors.
- "Only the fields marked required block registration" is not quite true for `rest`: an unset `${VAR}` in `headers` or `url` makes `build_camera_object()` return `None` (deliberate fail fast in `registry.py`), so the camera is absent from the registry instead of registering and misbehaving.

**3. Implicit contract details**

- The `start_zoom_focus` row: the zoom reset after a stream is best effort, wrapped in a broad `except` in `routes_stream.py` and always returns 200. Only `/control/zoom/{camera_ip}/{level}` gives the 400.
- The `focus_finder()` contract (polls `should_abort()`, raises `FocusAbortedError`) is contradicted by one of the two files the doc says to copy: the Linovision implementation is a stub that discards the callback and never raises, so the stream priority abort doesn't work there. Worth flagging which model file to trust.
- Path anchoring: line 17 declares paths relative to `pyro_camera_api/pyro_camera_api/`, but `setup_presets/` and `pyro_camera_api/tests/` only resolve from the repo root.

Group 1 is a quick fix, groups 2 and 3 are a few sentences each. After that this is good to go.

MateoLostanlen (approved):
thanks for the edit

PR #412, opened 2026-09-08
fix(deps): declare huggingface_hub in pyro-predictor and drop the 0.23.1 pin

MateoLostanlen:
Dependency cleanup around `huggingface_hub`.

- `pyro_predictor` now declares `huggingface_hub` (it imports it in `vision.py` but relied on `pyroengine` pulling it in), and `pyroengine` drops its unused `==0.23.1` pin that blocked every consumer needing a recent version (pyro-train for instance).
- `huggingface_hub` upgraded to 1.30.0 (latest), floor set to `>=1.30.0`. The only call site, `hf_hub_download(repo_id=, filename=, local_dir=)`, is unchanged in 1.x.
- `verbose=False` now also silences hf_hub's httpx request logs at runtime, and the test filters caplog to the `pyro_predictor` logger.
- `requirements.txt` regenerated with `make lock`: `opencv-python` stays stripped as since #338, and the redundant `--no-emit-package pyroclient` flag is gone.

Stacked on #414 (bookworm base images, Reolink legacy TLS): merge that one first. This diff only holds the dependency changes.


fe51 (changes requested):
Hi @MateoLostanlen , thanks for the PR, it looks good overall — `mypy` clean, sensors tests pass, and dropping the `huggingface_hub` pin is safe (I grepped: `pyroengine` never imports it, and `hf_hub_download(repo_id=, filename=, local_dir=)` is unchanged in 1.x).

**Main question: was `requirements.txt` regenerated by hand rather than with `make lock`?** The header comment records only the `uv export` half of the recipe — `Makefile:27` also greps:

```make
grep -v 'git+' ... | grep -v '^opencv-python==' > requirements.txt
```

Two consequences, unless you hit something that made `make lock` unusable:

1. **`opencv-python==4.13.0.92` is back** — a 46.7 MB aarch64 wheel that #338 deliberately stripped for 4G deploys. It now also sits next to `opencv-python-headless`; both install `cv2/`, so whichever `uv` writes last wins. Was that reintroduction intentional?
2. **`--no-emit-package pyroclient` looks redundant** — `grep -v 'git+'` already covers it, and since the Makefile wasn't touched the next `make lock` reverts this commit. Did you see a case where it still leaked through?

If not, `make lock` should fix both. (Heads-up: the orphan `# via ncnn` comments on `develop` are correct Makefile output, not a bug.)

---

Non-blocking:

3. **The new TLS test never runs.** CI and `make test` only run `tests/`; from the repo root it fails to import (`No module named 'pyro_camera_api.camera'`) and needs `PYTHONPATH=pyro_camera_api`. Pre-existing for all 7 files in that dir — do you want to wire the directory into CI here, or open a separate issue?

4. **Was the caplog filter meant as the fix for `verbose=False`, or just to unblock CI?** `vision.py:28` sets a root handler at INFO and `verbose=False` only lowers the `pyro_predictor` logger, so hf_hub's httpx lines still print at runtime. If suppressing them is in scope, one line in `__init__` does it:
   ```python
   logging.getLogger("httpx").setLevel(logging.WARNING)
   ```

5. **Was the module-level `requests.Session` deliberate?** No credential bleed (auth is in the URL, not session state), but it means one connection pool for every camera, shared across patrol/stuck-check threads and FastAPI's threadpool, and keep-alive now persists where each call used to get a fresh socket. With `max_retries=0` in requests, a socket the camera drops surfaces as `ConnectionError` — old Reolink firmware is exactly the population that does that. Patrol catches it, so worst case is intermittent skipped poses. `self._session` in `__init__` would remove all three, if there's no reason to share it.

6. **Was the `hf-xet` dependency considered?** hf_hub 1.x pulls it on aarch64 and downloads via `transfer.xethub.hf.co` instead of `cdn-lfs.huggingface.co` — a new host for egress-restricted sites. If it wasn't planned, `HF_HUB_DISABLE_XET=1` is the escape hatch.

MateoLostanlen:
Thanks @fe51, good catches all around.

1 and 2: you were right, the export was done by hand. Regenerated with `make lock`, so `opencv-python` is stripped again and the redundant `--no-emit-package pyroclient` is gone from the header.

3 and 5: the TLS test and the Reolink session moved to #414, which we will merge first (this branch now includes it, so the diff will collapse to the dependency changes once it lands). On the session, #414 sends `Connection: close` on every request, so no socket persists between calls. The `pyro_camera_api` test suite is now wired into CI and `make test` directly in #414, all 40 tests pass.

4: the caplog filter was just to unblock CI. Runtime fix added: `verbose=False` now also sets the `httpx` logger to WARNING.

6: hf-xet was not planned, good catch. Not blocking for us though: our sites do not filter this egress, except potentially the Ardèche one, and that pipe needs fixing anyway so we will handle it during the redeploy. We will just take the new default, and `HF_HUB_DISABLE_XET=1` stays the escape hatch if a site ever misbehaves.

PR #414, opened 2026-09-08
fix(camera): restore legacy TLS ciphers for old Reolink firmware, drop python 3.9 pin

MateoLostanlen:
## Context

pyro-camera-api was stuck on Python 3.9 (#237) because one station runs Reolink cameras on old firmware whose HTTPS handshake fails with Python 3.10+.

Diagnosis on the station (RLC-810A, firmware v3.1.0.764_22022803): the camera negotiates TLS 1.2 with secure renegotiation, but only offers RSA key exchange ciphers (AES128-GCM-SHA256, no ECDHE). Python 3.10 removed all non PFS ciphers from its default list, so the handshake fails. Restoring OpenSSL's own cipher list fixes it, verified live against the camera with Python 3.11 on the Pi.

## Changes

- Add a `_LegacyTLSAdapter` (requests HTTPAdapter that restores OpenSSL's default cipher list, so RSA key exchange is offered again) in the Reolink adapter and in `pyroengine/sensors.py`; all camera calls now go through a session mounted with it. Same approach as `reolink_aio` used by Home Assistant. No security regression: these calls already used `verify=False`, min protocol stays TLS 1.2, and a camera that offers ECDHE still gets ECDHE.
- The session sets `trust_env = False` (camera IPs are local, no proxy) and `Connection: close` so the pool never hands back a socket the camera dropped while idle, keeping the per-call behaviour of `requests.get`.
- Bump pyro-camera-api Docker image from `python:3.9.16-slim` to `python:3.11-slim-bookworm`, aligned with the engine image, and `requires-python` to `>=3.11`; regenerate the camera-api lock and requirements for 3.11.
- Now that 3.9 is gone, `pose_azimuths.py` uses `zip(..., strict=True)` again instead of the `# noqa: B905` workaround.
- Move both Docker base images off EOL bullseye to bookworm (bullseye's security Release file expired for good in August 2026, which was breaking every image build).
- Retarget `tests/test_sensors.py` mocks to the new session; add a test asserting the HTTPS session offers the RSA cipher these firmwares need.
- Run the `pyro_camera_api` test suite in CI and in `make test`: it had no runner until now, so the new TLS test would not have been executed.


fe51 (commented):
Hi @MateoLostanlen , thanks for this important PR 

Cipher part looks fine to me. Checked it on 3.11: `set_ciphers("DEFAULT")` goes from 17 to 60 ciphers,
min protocol stays TLS 1.2, no RC4/3DES/NULL, and we were already on `verify=False`. A camera that offers
ECDHE still gets ECDHE. So no worry for the other stations on that point.

Four questions though:

1. **The shared `_session` worries me more than the ciphers.** We went from a fresh connection per call
   to a pooled one with keep-alive, for every station. Reolink CGI drops idle connections and
   `HTTPAdapter` defaults to `max_retries=0` — won't we start seeing random `ConnectionError` in the
   patrol loops? Did you test a station over a few hours, or should we just add
   `_session.headers["Connection"] = "close"` and keep the old behaviour?

2. Same session is shared by the patrol threads and the FastAPI handlers. `requests.Session` isn't
   thread-safe (cookie jar). Probably harmless with Reolink, but is it intentional?

3. **3.9 → 3.11 changes more than the TLS.** `requirements.txt` has `python_full_version` markers, so
   the image now pulls requests 2.33.1, protobuf 7, websockets 16... plus bookworm (ffmpeg 4.3 → 5.1).
   None of that has ever run on a station. Can we soak it on one *non-legacy* site before the fleet?
   (I checked the ffmpeg flags, `-vsync 1` still works on 5.1, so that part should be ok.)

4. `pyro_camera_api/pyproject.toml` still says `requires-python = ">=3.9"` but `pose_azimuths.py` now
   uses `zip(..., strict=True)` (3.10+). Bump it to `>=3.11`?

Minor: the PR description still mentions `SECLEVEL=1` and legacy renegotiation, which 0511a4e removed —
worth fixing, ops will read that. And `pyro_camera_api/tests/` isn't in `make test` / CI (`pytest tests/`),
so the new `test_reolink_tls.py` never actually runs.

MateoLostanlen:
Thanks @fe51 , I did the updates !  I'll deploy to Brison and Saint-peray and test, especially the live stream, before deploying all fleet

